PR #5, opened 2023-03-07
Adding span to CETL

thirtytwobits:
The first polyfill class in CETL, span. This commit also establishes some of the infrastructure needed for cetlvast and provides full cetlvast coverage of the new span type.

Reviewers: concentrate on span.h and test_span.cpp. Comments on the cmake stuff are welcome but not required.


thirtytwobits:
Also, CI isn't setup yet. Once span is in place I'll setup our full suite on github actions.

include/cetl/cetl.h
```diff
@@ -0,0 +1,41 @@
+/// @file
+/// CETL common header.
+///
+/// Keep this very spare. CETL's desire is to adapt to future C++ standards
+/// and too many CETL-specific definitinos makes it difficult for users to switch off of CETL in the
```
emrainey, line 5: definitions

include/cetl/span.h
```diff
@@ -1 +1,772 @@
-// TODO: span for C++14.
\ No newline at end of file
+/// @file
+/// Defines a generic span type
```
emrainey, line 2: generic or standard's matching?

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    /// The value of Extent for this template instantiation.
+    static constexpr std::size_t extent = Extent;
+
+    // +----------------------------------------------------------------------+
+    // | Member Types
+    // +----------------------------------------------------------------------+
+    using element_type     = T;                                 //!< The element type over which the span operates
+    using value_type       = typename std::remove_cv<T>::type;  //!< The non-cv type of elements
+    using pointer          = T*;                                //!< The element pointer type
+    using const_pointer    = const T*;                          //!< A constant pointer to the element type
```
emrainey, line 52:
```suggestion
    using pointer_const    = T const*;                          //!< A pointer to the constant element type
```

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    using reference        = T&;                                //!< A reference type to the element
+    using const_reference  = const T&;                          //!< A constant reference type to the element
```
emrainey, line 54:
```suggestion
    using reference_const  = T const&;                          //!< A reference to the constant element type
```

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    /// @param first The first element in the span.
+    /// @param count The number of elements in the span.
+    /// @see std::span::span()
+    explicit constexpr span(iterator first, size_type count)
+        : data_{first}
+    {
+        (void) count;  // Count isn't actually used in static spans.
+        CETL_DEBUG_ASSERT(count == extent,
+                          "Undefined behavior: Constructing a fixed span where the Extent parameter is different from "
+                          "the count passed into this constructor.");
```
emrainey, line 93: This is about all you can do for static spans but people will complain. Also this will be problematic to detect at runtime on embedded targets (as me how I know).

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    /// @param first    The first element in the span.
+    /// @param end      The element after the last element in the span.
+    /// @see std::span::span()
+    template <typename EndType,
+              typename std::enable_if<!std::is_convertible<EndType, std::size_t>::value, bool>::type = true>
+    explicit constexpr span(pointer first, EndType end)
+        : data_(first)
+    {
+        (void) end;  // Last isn't actually used in static spans.
+        CETL_DEBUG_ASSERT(std::distance(first, end) == Extent, "Iterator range does not equal Extent.");
```
emrainey, line 109: This is about all you can do for static spans but people will complain. Also this will be problematic to detect at runtime on embedded targets (as me how I know).

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+
+    ///
+    /// Reference to an element in the span.
+    /// @param idx  The 0-based index in the span. The behavior of this method is
+    ///             undefined if idx is >= to size.
+    /// @return A reference to an element.
+    /// @see std::span::operator[]()
+    constexpr reference operator[](size_type idx) const
+    {
+        CETL_DEBUG_ASSERT((idx < Extent), "Indexing outside of the span is undefined.");
```
emrainey, line 213: This is about all you can do for static spans except wrapping the index. Can this span be empty? (nullptr, zero size?)

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    /// @see std::span::operator=()
+    constexpr span& operator=(const span& rhs) noexcept = default;
+
+    ///
+    /// Reference to an element in the span.
+    /// @param idx  The 0-based index in the span. The behavior of this method is
+    ///             undefined if idx is >= to size.
+    /// @return A reference to an element.
+    /// @see std::span::operator[]()
+    constexpr reference operator[](size_type idx) const
```
emrainey, line 211: since the method is `const`, shouldn't this be the `reference_const`? i.e. do we treat the data as part of the span (even through it not contained in the object proper)?

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    ///@}
+    // +----------------------------------------------------------------------+
+    ///@{ @name Observers
+    // +----------------------------------------------------------------------+
+
+    ///
+    /// If the span has a zero size or not.
+    /// @return true if the span size is 0 where "size" is the same as span::extent
+    /// for this specialization.
+    [[nodiscard]] constexpr bool empty() const noexcept
```
emrainey, line 317: Was wondering when I'd see the first `[[nodiscard]]`

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+
+    /// Create a new span `Count` elements from the last item of the current span to its end.
+    /// @tparam Count   Number of elements for the sub-span.
+    /// @return A new span with the static extent `Count`.
+    /// @see std::span::last
+    template <size_type Count>
+    constexpr span<element_type, Count> last() const
+    {
+        static_assert(Count <= extent, "subviews beyond the size of the span's view are not allowed.");
+        return span<element_type, Count>{(data_ + extent) - Count, Count};
```
emrainey, line 373: would prefer index math over pointer math.

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+
+    /// Create a new span `count` elements from the last item of the current span to its end.
+    /// @param count   Number of elements for the sub-span. The behavior of this method is undefined where
+    ///                `count` is greater than the current span's size.
+    /// @return A new span with a size of `count` and a dynamic extent.
+    /// @see std::span::last
+    constexpr span<element_type, cetl::dynamic_extent> last(size_type count) const
+    {
+        CETL_DEBUG_ASSERT(count <= extent, "Dynamic subviews beyond the size of the span's view are undefined.");
+        return span<element_type, cetl::dynamic_extent>((data_ + extent) - count, count);
```
emrainey, line 384: prefer index math over pointer math

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+
+    /// Create a new span `Offset` elements from the start of the current span and for `Count` elements.
+    /// @tparam Offset  Number of elements from the start of the current span for the subspan.
+    /// @tparam Count   The number of elements from the Offset to include in the subspan. If this value is
+    ///                 `cetl::dynamic_extent` then the Count is `size() - Offset`.
+    /// @return A new span where the `Extent` is `Count` if this was not `cetl::dynamic_extent` otherwise the
+    ///         new span's `Extent` is this span's `Extent` minus `Offset`.
+    /// @see std::span::subspan()
+    template <size_type Offset, size_type Count = cetl::dynamic_extent>
+    constexpr auto subspan() const -> span<element_type, (Count != cetl::dynamic_extent) ? Count : Extent - Offset>
```
emrainey, line 395: Since this is suppose to be C++20 everywhere, should the -> notion be more common? I'm not sure I like it but consistency is nice.

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    {
+        static_assert(Offset <= extent, "subspan Offsets > extent are ill-formed.");
+        static_assert((Count == cetl::dynamic_extent) || (Count <= extent - Offset),
+                      "subspan Count argument is ill-formed");
+        return span < element_type,
```
emrainey, line 400: this looks like a comma operator with a comparisons.  Not sure if there's a better way to write it. maybe just a comment?

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    /// The value of Extent for this template instantiation.
+    static constexpr std::size_t extent = cetl::dynamic_extent;
+
+    // +----------------------------------------------------------------------+
+    // | Member Types
+    // +----------------------------------------------------------------------+
+    using element_type     = T;                                 //!< The element type over which the span operates
+    using value_type       = typename std::remove_cv<T>::type;  //!< The non-cv type of elements
+    using pointer          = T*;                                //!< The element pointer type
+    using const_pointer    = const T*;                          //!< A constant pointer to the element type
```
emrainey, line 453:
```suggestion
    using pointer_const    = T const*;                          //!< A pointer to the constant element type
```

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    using reference        = T&;                                //!< A reference type to the element
+    using const_reference  = const T&;                          //!< A constant reference type to the element
```
emrainey, line 455:
```suggestion
    using reference_const  = T const&;                          //!< A reference to the constant element type
```

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    // +----------------------------------------------------------------------+
+    /// @copydoc span::operator=()
+    ///
+    constexpr span& operator=(const span& rhs) noexcept = default;
+
+    /// @copydoc span::operator[]()
+    ///
+    constexpr reference operator[](size_type idx) const
+    {
+        CETL_DEBUG_ASSERT((idx < size_), "Indexing outside of the span is undefined.");
```
emrainey, line 598: also nullptr, zero size

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    // +----------------------------------------------------------------------+
+    ///@{ @name Operators
+    // +----------------------------------------------------------------------+
+    /// @copydoc span::operator=()
+    ///
+    constexpr span& operator=(const span& rhs) noexcept = default;
+
+    /// @copydoc span::operator[]()
+    ///
+    constexpr reference operator[](size_type idx) const
```
emrainey, line 596: should be treat the data referenced in the span as const too?

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    ///
+    /// Returns if the span has a zero size or not.
+    /// @return true if the span size is 0.
+    [[nodiscard]] constexpr bool empty() const noexcept
+    {
+        return (size_ == 0);
+    }
+
+    ///
+    /// Returns the size of the span.
```
emrainey, line 671: it irks me that they didn't call this count() in the standard.

include/cetl/span.h
```diff
@@ -1 +1,772 @@
...
+    /// @return The number of elements in the span.
+    /// @see std::span::size
+    constexpr size_type size() const noexcept
+    {
+        return size_;
+    }
+
+    /// @copydoc span::size_bytes()
+    ///
+    constexpr size_type size_bytes() const noexcept
```
emrainey, line 681: For this exact reason. Could have had `size` and `count` mean two different things consistently, but no, now we have to qualify size instead.

cetlvast/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+
+if(NOT DEFINED CETL_ENABLE_DEBUG_ASSERT)
+    message(STATUS "CETL_ENABLE_DEBUG_ASSERT was not defined. Setting to off as a default")
+    set(CETL_ENABLE_DEBUG_ASSERT OFF)
+endif()
+
+if(NOT DEFINED CETLVAST_ENABLE_COVERAGE)
+    message(STATUS "CETLVAST_ENABLE_COVERAGE was not defined. Coverage will not be enabled.")
+    set(CETLVAST_ENABLE_COVERAGE OFF)
+endif()
```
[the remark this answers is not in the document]
thirtytwobits: I've never used option. Let me try this...

cetlvast/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+    message(STATUS "CETLVAST_TESTBUILD was not defined. Setting to \"native tests\" as a default")
+    set(CETLVAST_TESTBUILD ${CMAKE_SOURCE_DIR}/cmake/suites/native.cmake)
+else()
+    set(CETLVAST_TESTBUILD ${CETLVAST_TESTBUILD})
+endif()
+
+if(NOT DEFINED CETL_ENABLE_DEBUG_ASSERT)
+    message(STATUS "CETL_ENABLE_DEBUG_ASSERT was not defined. Setting to off as a default")
+    set(CETL_ENABLE_DEBUG_ASSERT OFF)
+endif()
```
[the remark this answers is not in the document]
thirtytwobits: Resolve this one as I'll work on all off these as part of the CETLVAST_ENABLE_COVERAGE conversation.

cetlvast/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+    message(STATUS "CMAKE_MODULE_PATH set to ${CMAKE_MODULE_PATH}")
+endif()
+
+if(DEFINED CETLVAST_CPP_STANDARD)
+    message(STATUS "CETLVAST_CPP_STANDARD was set to C++${CETLVAST_CPP_STANDARD} by the CETLVAST_CPP_STANDARD variable.")
+else()
+    #
+    # Tell cmake where to find our custom scripts.
+    #
+    set(CETLVAST_CPP_STANDARD "20")
```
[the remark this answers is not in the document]
thirtytwobits: How so? The main point of CETL is to provide forwards compatibility with C++17, 20, 23, and beyond.

cetlvast/cmake/compiler_flag_sets/native.cmake
```diff
@@ -0,0 +1,40 @@
...
+        message(STATUS "CETL_ENABLE_DEBUG_ASSERT will be 1 enabling debug assert() in cetl code.")
+        list(APPEND C_FLAG_SET
+            "-DCETL_ENABLE_DEBUG_ASSERT=1")
+    endif()
+endif()
+
+list(APPEND CXX_FLAG_SET ${C_FLAG_SET})
+list(APPEND ASM_FLAG_SET ${C_FLAG_SET})
+
+list(APPEND CXX_FLAG_SET)
```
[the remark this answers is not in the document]
thirtytwobits: oh weird. Fixed.

include/cetl/span.h
thirtytwobits: Why this change? This is taken from the C++ standard directly:
image

include/cetl/span.h
thirtytwobits: So, this is weird, but the standard defines c[begin/end] and cr[begin/end] but neither gcc nor clang actually implement this nor does cppreference.com list it. Otherwise, this operator is defined per the standard and I don't see any const reference version in it nor in cppreference.com.

include/cetl/span.h
thirtytwobits: sorry

include/cetl/span.h
thirtytwobits: sorry

include/cetl/cetl.h
thirtytwobits: no, definitinos is the Italian with a precise meaning of "this is all bullshit" :) (I'll fix)

include/cetl/span.h
thirtytwobits: standards matching. Will update.

include/cetl/span.h
thirtytwobits: same as before. I'll resolve this conversation and we'll talk about your proposal versus the C++ standard up there.

include/cetl/span.h
thirtytwobits: same as before. I'll resolve this conversation and we'll talk about your proposal versus the C++ standard up there.

include/cetl/span.h
thirtytwobits: details?

include/cetl/span.h
thirtytwobits: Yes. If you define the Extent as 0 the data_ pointer is null. I should expand this to be: `(Extent > 0) && (idx < Extent)`

include/cetl/span.h
thirtytwobits: Not according to the standard. That said, it still isn't writable memory even if the reference isn't marked as const:

https://godbolt.org/z/KhPKqa37d

include/cetl/span.h
thirtytwobits: Yeah, this is a smart move on LLVM's part (not sure if gcc followed suit). A lot of people think this is a verb. This makes it more probable that the compiler warning about not consuming the result will clue the user in.

include/cetl/span.h
thirtytwobits: This is the only way I know how to implement this because the return type depends on the template parameters. Nowhere else, in this class, is this true.

cetlvast/CMakeLists.txt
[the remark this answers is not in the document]
emrainey: Will it be compatible w/ 14 though?

include/cetl/span.h
emrainey: Even the standard doesn't logically make sense as the reference isn't constant, the object it references is.

include/cetl/span.h
emrainey: Not your fault. I probably need to write my own standard library, I'll call it the `grumpy template library` or `gtl::`

include/cetl/span.h
emrainey: Maybe I'll call my TL, `curmudgeon's template library` instead.

cetlvast/CMakeLists.txt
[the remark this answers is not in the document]
thirtytwobits: Oh, sure. Of course that is the point. When I setup the CI I'll have jobs running at C++14, 17, 20, and 23. When a polyfill type like span appears the parameterized tests will be able to compare them at the same time.

include/cetl/span.h
thirtytwobits: I'll be your first customer.

include/cetl/span.h
thirtytwobits: `ykgoml::` perhaps? (You Kids! Get Off My Lawn!!)

include/cetl/span.h
pavel-kirienko: nodiscard is not supported in C++14 so should not be used here unless made conditional on the standard version

include/cetl/span.h
pavel-kirienko: both are prohibited by AUTOSAR C++ though

include/cetl/span.h
pavel-kirienko: The way the code is formatted makes it hard to follow. Was it clang-format? If so, a newer version may fix this as there've been many issues of this kind before LLVM 15.

include/cetl/span.h
pavel-kirienko: Why can't you move the return type declaration to the left? It is not dependent on the runtime arguments since there are none

include/cetl/span.h
thirtytwobits: it is clang-format from the ts20.4.1 container. We'd need to update it there.

include/cetl/span.h
thirtytwobits: ah crud. You're right. I think it's best to omit this since it doesn't affect the ABI in any way. Or should we add a CETL_NODISCARD macro? But macros are bad.

thirtytwobits:
Updated the C++14 issues but I haven't addressed everything in the review yet.

include/cetl/span.h
thirtytwobits: I was using the wrong clang-format. I just pushed a change that is formatted using the Toolshed container version. Let me know if this is what you expected.

pavel-kirienko: It's still the same. Maybe it's time to update the container? On the other hand, this is just formatting, it doesn't affect validity so this issue is a very low-priority one. We can upgrade the container later.

cetlvast/CMakeLists.txt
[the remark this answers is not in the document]
thirtytwobits: neat!

include/cetl/span.h
thirtytwobits: I'm proposing a slightly different approach where we do use macros but we explicitly put the feature testing in the header. Continuing this pattern will prevent cetl.h from becoming a uber portability layer and it lets the user of these types see, very clearly, what adaptations our types make based on the standard in use. Let me know what you think.

include/cetl/span.h
thirtytwobits: https://github.com/OpenCyphal/docker_toolchains/issues/25

cetlvast/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+include(${CETLVAST_FLAG_SET})
+
+# +---------------------------------------------------------------------------+
+# | EXTERNAL DEPENDENCIES
+# +---------------------------------------------------------------------------+
+
+#
+# googletest (and googlemock) external project.
+#
+find_package(gtest REQUIRED)
```
thirtytwobits, line 117: eh, I should move this into the unittest suite.cmake. It's not used for any other test suite.

cetlvast/cmake/modules/Finddocs.cmake
```diff
@@ -0,0 +1,105 @@
+#
+# This treats the doxygen build for CETL as a standalone program. In
+# reality the docs target is a doxygen build configured just for this project.
+#
+
+
+find_package(Doxygen QUIET)
```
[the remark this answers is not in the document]
thirtytwobits: I need to change this to required. Good catch.

include/cetl/span.h
pavel-kirienko: I am not sure I understand your proposal in its entirety but it sure sounds great.

include/cetl/span.h
emrainey: misconfiguring the compile time size and the runtime size, causing the assert can be difficult to track down and ultimately a weird problem to choose to allow. My personal preference is to make `span` an interface which is either implemented by a FixedSpan or a DynamicSpan.

cetlvast/CMakeLists.txt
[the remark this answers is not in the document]
emrainey: May look something like this
```cmake
option(USE_THUMB_MODE "Enable Thumb Instruction Set" ON)
if(USE_THUMB_MODE)
    add_compile_options(-mthumb -mno-thumb-interwork)
    message(STATUS "Using Thumb instruction set")
else()
    message(STATUS "Using ARM instruction set")
endif()
```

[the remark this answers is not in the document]
emrainey: I think options can surface into Cmake GUI tools as configuration items.

include/cetl/span.h
emrainey: index math is more clear and in a single "domain" instead of having an implied `*sizeof(X)`

include/cetl/span.h
emrainey: I just mean that the  `span < element_type`  looks like testing if `span` is `<` than `element_type`

include/cetl/span.h
emrainey: The pointer is not const, the thing it points to is.

include/cetl/span.h
pavel-kirienko: That's what old clang-format does to templates, yes.

include/cetl/span.h
thirtytwobits: see my latest push

include/cetl/span.h
thirtytwobits: Based on the specification this constructor:
image
given the required preconditions:
image
and that `count` has to be evaluated at runtime, I'm not sure there is any other recourse here.

cetlvast/CMakeLists.txt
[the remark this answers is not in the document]
thirtytwobits: wow, and Nobody Everston uses Cmake GUI all the time too!

include/cetl/span.h
thirtytwobits: Resolving this as the implementation is per ISO/IEC specification
image

include/cetl/span.h
thirtytwobits: Resolving this since this is per ISO/IEE
image

include/cetl/span.h
thirtytwobits: We good with the latest @pavel-kirienko ?

include/cetl/span.h
thirtytwobits: Huh. you're right. Somehow I convinced myself that this wouldn't work but...it totally does.

include/cetl/span.h
thirtytwobits: This is per ISO/IEC. Sorry
image

include/cetl/cetl.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef CETL_H_INCLUDED
+#define CETL_H_INCLUDED
+
+//
+// @note CETL_ENABLE_DEBUG_ASSERT
+// Define as 1 to enable library assertions. Enabling this in production code is
+// <em>strongly</em> discouraged.
+//
+#ifndef CETL_ENABLE_DEBUG_ASSERT
+#    ifdef DEBUG
```
pavel-kirienko, line 20: Surely you mean `NDEBUG`?

include/cetl/span.h
pavel-kirienko: It works, but I suspect it might get cumbersome further down the road when we add more optinoal capabilities like `[[noinline]]` etc. I am fine with either approach but all other things equal I would incline towards `CETL_NODISCARD`.

include/cetl/cetl.h
thirtytwobits: man, this macro section is all wrong. Good catch.

include/cetl/span.h
thirtytwobits: Resolving this. Will finish this discussion in the other conversation.

include/cetl/span.h
thirtytwobits: Okay.

thirtytwobits:
Okay. we good with this now?

pavel-kirienko (approved):

Issue #16, opened 2023-03-09
automatic polyfill header


thirtytwobits:
I'm only 90% sure this will work but we should provide a header like "cetl_polyfill.h" that uses C++20 feature testing (https://en.cppreference.com/w/cpp/feature_test) to include CETL types when they are absent from a given environment. This would support compilers that support the core C++20 language but which do not provide a full-featured standard library.

Not sure of the value of this so this should be a low priority task.

pavel-kirienko:
there definitely is value in this

PR #18, opened 2023-03-18
Fixing #16 and introducing polyfill namespaces.

thirtytwobits:


include/cetl/cetl.hpp
```diff
@@ -0,0 +1,81 @@
...
+/// @{
+
+/// @def CetlCppStandard14
+/// Provides the proper value to test against `__cplusplus` for c++14.
+/// ```
+/// #if __cplusplus >= CetlCppStandard14
+/// #include <something_from_cpp_14>
+/// #endif
+/// ```
+#define CetlCppStandard14 201402L
```
pavel-kirienko, line 57: Maybe I missed something, why did you change these from the screaming snake case?

pavel-kirienko (approved):

include/cetl/pf20/cetlpf.hpp
```diff
@@ -0,0 +1,60 @@
...
+/// ```
+///
+
+#ifndef CETL_PF20_H_INCLUDED
+#define CETL_PF20_H_INCLUDED
+
+#include "cetl/cetl.hpp"
+
+#if (__cplusplus >= CetlCppStandard20)
+#    include <span>
```
emrainey, line 42: Is this std:: style?

include/cetl/pf20/cetlpf.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#else
+#    include "cetl/pf20/span.hpp"
```
emrainey, line 45: here too

include/cetl/cetl.hpp
emrainey: I'm confused about that too. I dislike the macros (they are avoidable but with some hassle) but if they remain, should be as they were..

thirtytwobits: Because, if we were to drop C++14 these would become inline variables. This future proofs the tokens for such a glorious future (which is not forthcoming this year nor next but perhaps someday).

include/cetl/pf20/cetlpf.hpp
thirtytwobits: Yeah. @pavel-kirienko , I'm getting the right clang-format style here right?

pavel-kirienko: Seems so.

include/cetl/cetl.hpp
pavel-kirienko: >Because, if we were to drop C++14 these would become inline variables.

Wait what? Are you saying that in C++17+ you can expand inline variables in preprocessor conditions?

thirtytwobits: Indeed: https://godbolt.org/z/c3v3Y8YT7

thirtytwobits: Yeah, I'm an idiot. I'll `CHANGE_THESE_BACK_SORRY`.

pavel-kirienko (approved):

emrainey (approved):

PR #19, opened 2023-03-21
more pro forma changes before graduation

thirtytwobits:


include/cetl/pf20/span.hpp
```diff
@@ -329,7 +331,7 @@ class span
     /// If the span has a zero size or not.
     /// @return true if the span size is 0 where "size" is the same as span::extent
     /// for this specialization.
-#if (__cplusplus >= CETL_CPP_STANDARD_17)
+#if (__cplusplus >= 201703L)
```
emrainey, line 334: why the number change to 17 #defines?

include/cetl/pf20/span.hpp
```diff
@@ -687,7 +689,7 @@ class span<T, dynamic_extent>
     ///
     /// Returns if the span has a zero size or not.
     /// @return true if the span size is 0.
-#if (__cplusplus >= CETL_CPP_STANDARD_17)
+#if (__cplusplus >= 201703L)
```
emrainey, line 692: link to the published C++ standards numbers would be useful, but I tried to find it myself and had difficulty. This might be the only definitive answer https://en.cppreference.com/w/cpp/preprocessor/replace

include/cetl/pf20/span.hpp
thirtytwobits: Since the inclusion of cetl.h asserts that this is a value that we expect to see and refuses to use anything like `201704L` etc I've decided to reduce the number of places within this type we need something named "CETL" to compile.

emrainey (approved):

emrainey (approved):

PR #23, opened 2023-04-04
Enabling CI

thirtytwobits:
* switching over to gcovr for coverage reporting
* switching to upload docs by directory
* enabling sonarqube upload

(good lord this was hard to get working!)


cetlvast/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: MIT
 #
 
-cmake_minimum_required(VERSION 3.16.0)
+cmake_minimum_required(VERSION 3.22.0)
 
 ###############################################################################
 # DON'T USE CMAKE!
```
pavel-kirienko, line 10: image

cetlvast/suites/docs/images/html/docs-ok-green.svg
pavel-kirienko, line 1: https://shields.io ?

sonar-project.properties
pavel-kirienko, line 1: The way it is currently configured, SonarCloud will treat both the test suite and the library itself identically. Maintaining the test suite to the same quality standard is usually impractical so I suggest modifying the configuration such that the test suite is excluded from Sonar analysis entirely. It is a bit tricky to set up with SonarCloud but is possible, here's an example: https://github.com/Zubax/kocherga/blob/69e2131d3a26807428f67dc2f823afd988da1bc7/.github/workflows/main.yml#L181-L196

pavel-kirienko (changes requested):

sonar-project.properties
thirtytwobits: Okay, good comment. I fixed this up and I included test execution reports since, without coverage data for the tests themselves, the only indication we actually ran the tests is in the report.

Also, totally forgot about `>` in yaml. Thanks!

pavel-kirienko (approved):

thirtytwobits:
I'm going to close this and open a new PR from a branch so we can make sure it actually works with sonarqube first...

PR #26, opened 2023-04-20
Adding polymorphic_allocator and memory_resource(s)

thirtytwobits:
Adding several types from std::pmr and some new memory_allocator implementations specific to CETL.

This is a preliminary PR to act as a design review. Obviously there's missing docs and tests that will be needed before this can be accepted.


cetlvast/include/cetlvast/helpers.hpp
```diff
@@ -12,8 +12,48 @@
...
+#    endif
+#endif
+
+namespace cetlvast
+{
+
+template <typename T>
+constexpr bool is_power_of_two(const T& value)
+{
+    return (value && !(value & (value - 1)));
```
pavel-kirienko, line 40: Do I understand correctly that the implicit cast to bool is allowed here only because it is not production code but test code, and the coding standard enforcement is relaxed here?

pavel-kirienko (commented):
It almost feels like I am a bit out of context on this change so it is a bit challenging to review sensibly. I see that there is a lot of attention given to memory management but how does it relate to the ongoing work on libcyphal is not entirely clear to me. I am not complaining but rather pointing out my limited ability to provide high-level feedback here.

I have a tangential question that's been on my mind once: does the C++17/20 standard library provide a way of managing unique pointers to objects allocated by memory resources that is better that manually instantiating a `std::unique_ptr<T, Deleter>` with a custom `Deleter` that calls `std::pmr::memory_resource::deallocate` (similar to your `MemoryBlockPointer`)? The problem with this approach is that it is unwieldy and dangerous, as it is too easy to accidentally use a regular smart pointer with the default deleter, which results in UB/heap corruption upon deletion.

Also, would it not be nice to provide a default `std::pmr::memory_resource` implementation for O1Heap as a C++17 extension? I suppose it would go into the O1Heap repository but it could be designed to be natively interoperable with CETL.

pavel-kirienko (commented):
I see that you introduced a completely custom unique pointer implementation. Would it not be sufficient to use `std::unique_ptr<T, Destroyer>`, where the destroyer is defined roughly as:

```c++
class Destroyer final
{
public:
    explicit Destroyer(memory_resource* const heap) : heap_(heap) {}

    template <typename T>
    void operator()(T* const obj) const
    {
        static_assert((sizeof(T) > 0) && (!std::is_void_v<T>), "incomplete type");
        if (obj != nullptr)
        {
            obj->~T();
            assert(heap_ != nullptr);
            heap_->deallocate(obj);
        }
    }

private:
    memory_resource* heap_;
};
```

thirtytwobits:
> I see that you introduced a completely custom unique pointer implementation. Would it not be sufficient to use `std::unique_ptr<T, Destroyer>`, where the destroyer is defined roughly as:
> 

I thought your objection was around this technique. I've misunderstood. This is, typically, how one handles these sorts of things but you seemed to imply this can be undone by certain transformations of unique_ptr instances? If you are okay with the idiomatic method I'll gladly switch to that.

pavel-kirienko:
>This is, typically, how one handles these sorts of things but you seemed to imply this can be undone by certain transformations of unique_ptr instances? If you are okay with the idiomatic method I'll gladly switch to that.

I am okay with the idiomatic method, which is not to say that I am happy with it. I thought that there could be a more off-the-shelf solution, like a custom smart pointer implementation in the `std::pmr` namespace (there is none), or something similar.

PR #27, opened 2023-05-02
Develop/vla

thirtytwobits:
Combines all the work on PMR with the addition of the VariableLengthArray


cetlvast/suites/docs/examples/example_05_array_memory_resource_alignment.cpp
```diff
@@ -0,0 +1,123 @@
...
+//![example_delegate]
+
+int main()
+{
+    {
+        static_assert(alignof(std::max_align_t) < 128, "Wow, what hardware are you running on?");
+//![example_0]
+        constexpr std::size_t                        BufferSizeBytes = 64;
+        cetl::pf17::byte                             buffer[BufferSizeBytes];
+        cetl::pmr::UnsynchronizedArrayMemoryResource resource{buffer, BufferSizeBytes};
```
skeetsaz, line 85: Upstream can't be null
```
cetl::pmr::UnsynchronizedArrayMemoryResource resource{buffer, BufferSizeBytes, cetl::pf17::
pmr::null_memory_resource()};
```

cetlvast/suites/docs/examples/example_05_array_memory_resource_alignment.cpp
```diff
@@ -0,0 +1,123 @@
...
+        std::cout << "Over-aligned attempt failed: " << r << std::endl;
+
+//![example_0]
+    }
+    {
+//![example_1]
+        // By over-provisioning the buffer you can now get the alignment you want:
+        constexpr std::size_t                        BufferSizeBytes = 64 + 128;
+        cetl::pf17::byte                             buffer[BufferSizeBytes];
+        cetl::pmr::UnsynchronizedArrayMemoryResource resource{buffer, BufferSizeBytes};
```
skeetsaz, line 112: Same here.  Upstream can't be null.

cetlvast/suites/docs/examples/example_05_array_memory_resource_array.cpp
```diff
@@ -0,0 +1,120 @@
...
+        static cetl::byte upstream_buffer[SmallMessageSizeBytes];
+        cetl::pmr::UnsynchronizedArrayMemoryResource<cetl::pmr::memory_resource>
+            cUpstreamResource{&upstream_buffer, SmallMessageSizeBytes, cetl::pmr::null_memory_resource()};
+        cetl::pmr::UnsynchronizedArrayMemoryResource<cetl::pmr::memory_resource> cResource{small_message_buffer_,
+                                                                                           SmallMessageSizeBytes,
+                                                                                           &cUpstreamResource};
+        cetl::pmr::polymorphic_allocator<std::uint64_t>                          cAlloc{&cResource};
+        Message                                                                  c{cAlloc};
+
+        // This time we won't reserve which should cause vector to do multiple allocations. We'll also insert
```
skeetsaz, line 104: Fix comment.  "We'll also insert a bunch more" is not true.

cetlvast/suites/docs/examples/example_05_variable_length_array_vs_vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+int main()
+{
+#if !__cpp_exceptions
+    {
+        //! [example_no_exceptions]
+
+        // Using the cetlpf.hpp header we can create a polymorphic allocator that is aliased to
+        // std::pmr::polymorphic_allocator when compiling with C++17 or newer and cetl::pf17::pmr::polymorphic_allocator
+        // when compiling with C++14.
+        cetl::pmr::polymorphic_allocator<int> alloc{cetl::pmr::new_delete_resource()};
```
skeetsaz, line 53: Bug:  Template type mismatch among alloc (int) and a/b (char).  Why does this compile?

cetlvast/suites/docs/examples/example_06_polymorphic_alloc_deleter.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include <string.h>
+
+class MyObject final
+{
+public:
+    MyObject(const char* name, std::size_t name_length)
+        : name_(nullptr)
+    {
+        name_ = static_cast<char*>(malloc(name_length + 1));
+        strncpy(name_, name, name_length);
```
skeetsaz, line 27: Might want to ensure name_ is null-terminated since strncpy doesn't guarantee it.

include/cetl/pf17/cetlpf.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+namespace cetl
+{
+namespace pmr
+{
+
+/// Automatic implementation of cetl::pmr::memory_resource that uses
+/// cetl::pmr::UnsynchronizedArrayMemoryResourceDelegate as the internal implementation.
+/// @tparam UpstreamResourceType The type of any upstream memory resource provided.
+template <typename UpstreamResourceType>
```
skeetsaz, line 109: It's a little weird that this requires a template parameter UpstreamResourceType.  Is this to keep it header-only?  And also I guess UnsynchronizedArrayMemoryResourceDelegate needs a way to refer to the type of the upstream but it can't since memory_resource is ultimately defined here in the polyfill header?

include/cetl/pmr/memory.hpp
```diff
@@ -0,0 +1,231 @@
...
+{
+
+/// RAII helper for memory_resource::allocate() and memory_resource::deallocate(). This type is designed to work with
+/// std::unique_ptr.
+///
+/// Example usage:
+///
+/// @snippet{trimleft} example_06_memory_resource_deleter.cpp example_usage
+///
+class MemoryResourceDeleter final
```
skeetsaz, line 32: Does a full class definition in the header violate the One Definition Rule?  Do you have to put "inline" or the methods for it to be correct in a header-only library?

include/cetl/pmr/memory.hpp
```diff
@@ -0,0 +1,231 @@
...
+{
+public:
+    using allocator  = PolymorphicDeallocator;
+    using value_type = typename PolymorphicDeallocator::value_type;
+    using unique_ptr = std::unique_ptr<value_type, cetl::pmr::PolymorphicDeleter<PolymorphicDeallocator>>;
+
+    PolymorphicDeleter() noexcept = delete;
+
+    PolymorphicDeleter(const PolymorphicDeallocator& alloc,
+                       std::size_t mem_size) noexcept(std::is_nothrow_copy_constructible<PolymorphicDeallocator>::value)
```
skeetsaz, line 140: "mem_size" seems misnamed since this is the number of objects not the number of bytes

skeetsaz, line 140: "mem_size" seems misnamed since this is the number of objects not the number of bytes

include/cetl/pmr/o1heap_memory_resource.hpp
```diff
@@ -0,0 +1,103 @@
+/// @file
+/// Defines cetl::pmr::O1HeapMemoryResource which is a std::pmr::memory_resource implemented in terms of Pavel
+/// Kirienko's [o1heap](https://github.com/pavel-kirienko/o1heap). If including this file you will need to either first
+/// include cetl/pf17/cetlpf.hpp or provide the memory_resource definition you want this class to use.
+/// You'll also need to provide an include path to o1heap.h and compile in o1heap.c where use use this type.
```
skeetsaz, line 5: Typo: "use use"

include/cetl/variable_length_array.hpp
```diff
@@ -0,0 +1,1710 @@
...
+    using allocator_type = typename std::allocator_traits<Allocator>::template rebind_alloc<T>;
+
+    /// Required constructor for std::uses_allocator protocol.
+    /// @param alloc Allocator to use for all allocations.
+    /// @param max_size_max Clamping value for the maximum size of this array. That is,
+    /// cetl::VariableLengthArray::max_size() will return
+    /// `std::min(max_size_max, std::allocator_traits<Allocator>::max_size(alloc))`
+    explicit constexpr VariableLengthArray(
+        const allocator_type& alloc,
+        size_type             max_size_max =
```
skeetsaz, line 102: These max_size_max calculations seem wrong.  Allocator max_size returns the maximum number of elements, not the maximum number of bytes (see https://en.cppreference.com/w/cpp/memory/allocator/max_size)

include/cetl/pmr/memory.hpp
pavel-kirienko: >Does a full class definition in the header violate the One Definition Rule?

No.

>Do you have to put "inline" or the methods for it to be correct in a header-only library?

Member functions defined inside the class body are always implicitly declared `inline`.

include/cetl/variable_length_array.hpp
thirtytwobits: Good catch. It's actually a bit more nuanced then that. The maxsize is actually the maximum diffval since you have to be able to iterate over all elements in a collection even if you can allocate more memory (theoretically, of course)

include/cetl/pmr/memory.hpp
thirtytwobits: fixed and I added more docs.

include/cetl/pmr/o1heap_memory_resource.hpp
thirtytwobits: fixed

include/cetl/variable_length_array.hpp
thirtytwobits: Actually, now that I look at this it is correct. So, the max_size_max value allows a user to set an artificial ceiling. For example:

```
auto foo = VariableLengthArray(std::allocator(), 24);
```

since std::allocator doesn't implement max_size you can set a max_size_max  to implement a DSDL bounded-variable-length array. That this max size default is too large is okay because we always correctly calculate the value in the max_size() method.

Also, max_size_bytes == max_size_objects where sizeof(T) == 1.

cetlvast/suites/docs/examples/example_05_array_memory_resource_alignment.cpp
thirtytwobits: I'm not following you here. There's a default argument that returns a non-null pointer. Am I not seeing something?

cetlvast/suites/docs/examples/example_05_array_memory_resource_alignment.cpp
thirtytwobits: same as above.

cetlvast/suites/docs/examples/example_05_array_memory_resource_array.cpp
thirtytwobits: Good catch. Did a bunch of cleanup of the example text here.

cetlvast/suites/docs/examples/example_05_variable_length_array_vs_vector.cpp
thirtytwobits: everything was signed and no implicit casts narrowed any types so, while it is wrong (and I fixed it), it would only waste memory (but might be more efficient for some memory resources on some platforms due to alignment).

cetlvast/suites/docs/examples/example_05_variable_length_array_vs_vector.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+        print_container(a);
+        print_container(b);
+
+        // So why not just use vector?
+        // The primary reason is std::vector has some edge cases where it cannot be used when exceptions are disabled.
+        // Here we are giving a vector and a VariableLengthArray the same allocator CETL pf17 allocator. This allocator
+        // returns nullptr when out of memory and exceptions are disabled.
+
+        cetl::pf17::pmr::polymorphic_allocator<int> bad_alloc{cetl::pf17::pmr::null_memory_resource()};
```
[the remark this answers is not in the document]
thirtytwobits: fixed

cetlvast/suites/docs/examples/example_06_polymorphic_alloc_deleter.cpp
thirtytwobits: fixed

include/cetl/pf17/cetlpf.hpp
thirtytwobits: That's right. The polyfill type and the inner, delegate support using std::pmr types or cetl::pmr types so this template parameter allows either to be generated from the same logic.

cetlvast/suites/docs/examples/example_05_array_memory_resource_alignment.cpp
skeetsaz: Oops, I copy/pasted from my local repo where I had put comments and was out of sync.  Looks like this is fixed now.  Ignore.

cetlvast/suites/docs/examples/example_05_array_memory_resource_alignment.cpp
skeetsaz: Ignore per previous comment.

include/cetl/variable_length_array.hpp
skeetsaz: Hmm, there is some subtly here.  max_size() on allocator, allocator_traits, and vector, all represent the theoretical maximum number n of T elements (regardless as to whether there is that much memory available.)

If we followed that model then VariableLengthArray::max_size() should also represent the theoretical max and the impl should just be numeric_limits<size_type>::max() / sizeof(value_type).

Instead of max_size_max, I think the concept we should expose is capacity_max, passing that as a constructor arg and having a capacity_max() method.  This keeps VariableLengthArray::max_size() semantically the same as allocator/allocator_traits/vector and exposes VLA's unique new property, that of being bounded, as a new API (which the vector API has no way to represent.)

What do you think?

PR #28, opened 2023-05-30
Adding VariableLengthArray and PMR namespace

thirtytwobits:
Providing a polyfill of much of the pmr namespace for C++14.

Adding VariableLengthArray as an allocator-aware vector. CETL's VLA differs from vector in three respects:
1. it supports the CETL-specific reallocation extension to PMR::memory_resource
2. it will tightly fit into a maximum size when elements are pushed without calling reserve first.
3. it avoids undefined behaviour when exceptions are disabled.

An unintended benefit: this works with C++17/AppleClang which does not support PMR.

## Reviewer Notes

As always, feel free to skip the vscode/cmake stuff. There was a major refactor of this as I learned more about cmake and I enabled sharing modules with dependent projects (i.e. libcyphal).

This is a huge change. Sorry. Circumstances prevented me from getting PMR in before VLA. If the review is too daunting I can split it up into three stages but it would take some time to get each building independently. One way to make this easier is to review everything under the PMR namespace first and ignore VLA to start with then come back to the VLA.

There are known missing comments, unittests, and CI fix-ups that I'll be working on in parallel.


include/cetl/variable_length_array.hpp
```diff
@@ -0,0 +1,2049 @@
...
+    }
+    constexpr static size_type bits2bytes(const size_type value) noexcept
+    {
+        return round8(value) / 8U;
+    }
+
+    constexpr size_type size_bits() const noexcept
+    {
+        CETL_DEBUG_ASSERT(last_byte_bit_fill_ <= 8, "last_byte_bit_fill_ is out of range.");
+        CETL_DEBUG_ASSERT(size_ <= capacity_, "size_ is out of range.");
```
thirtytwobits, line 2034: add assert that if size_ > 0 last_byte_bit_fill_ is also > 0.

.vscode/open_docs_mac.sh
```diff
@@ -0,0 +1,3 @@
+
+# source me from the root directory on mac to open vscode generated docs in browser
```
thirtytwobits, line 2: whoops. Remove this file.

pavel-kirienko:
Is there some place on the web where the docs could be deployed for review?

thirtytwobits:
> Is there some place on the web where the docs could be deployed for review?

I fixed the CI enough that the latest run uploaded an archive of the docs.

Also: The CI revealed that std::vector can refuse to shrink_to_fit when exceptions are disabled. I'll need to fixup the tests to account for this.

thirtytwobits:
>  clang-format-check.py when the same could be achieved in fewer lines of code via a custom cmake target 

Do tell! I'd love to get rid of this if I could but I didn't want to try to parse xml in cmake.

thirtytwobits:
> would strongly prefer to have VariableLengthArray::at(std::size_t) defined similarly to...

Sorry, but I'm not going to put a production `abort()` into the code. You can implement an external helper:

```
template<typename VLAType>
const typename VLAType::value_type& at(const VLAType& container, std::size_t pos)
{
    if (pos >= container.size())
    {
#if __cpp_exceptions
        throw std::out_of_range("VariableLengthArray::at");
#else
        std::abort();
#endif
    }
    return container[pos];
}

```

pavel-kirienko:
I normally use https://github.com/marketplace/actions/clang-format-lint together with a `format` build target that invokes clang-format when necessary, also there are IDE integrations for on-the-fly reformatting. If you want to keep the formatting check strictly a part of CMake project, then perhaps invoking clang-format with `clang-format --dry-run --Werror <paths...>` is what you need?

pavel-kirienko:
>Sorry, but I'm not going to put a production abort() into the code. You can implement an external helper:

How about the old `at_or_null`? For the bool specialization it would become:

```c++
std::optional<reference> at_or_null(const size_type);
std::optional<const_reference> at_or_null(const size_type) const;
```

thirtytwobits:
> > Sorry, but I'm not going to put a production abort() into the code. You can implement an external helper:
> 
> How about the old `at_or_null`? For the bool specialization it would become:
> 
> ```c++
> std::optional<reference> at_or_null(const size_type);
> std::optional<const_reference> at_or_null(const size_type) const;
> ```

Yes. I was planning on doing that when I added optional to CETL.

include/cetl/pmr/memory.hpp
```diff
@@ -0,0 +1,245 @@
...
+        }
+        alloc_.deallocate(p, obj_count_);
+    }
+
+private:
+    PolymorphicDeallocator alloc_;
+    std::size_t            obj_count_;
+};
+
+/// Factory helper for creating objects with polymorphic allocators using proper RAII semantics.
```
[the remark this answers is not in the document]
thirtytwobits: Added #29. I think this is completely additive so let's do that in another change.

cetlvast/cmake/modules/Findgoogletest.cmake
```diff
@@ -0,0 +1,408 @@
...
+# cmake 3.21 and earlier.  This workaround may not be needed after 3.24.
+get_property(googletest_POPULATED GLOBAL PROPERTY googletest_POPULATED)
+
+if(NOT googletest_POPULATED)
+
+    if (NOT FETCHCONTENT_SOURCE_DIR_googletest)
+        set(FETCHCONTENT_SOURCE_DIR_googletest ${CETLVAST_EXTERNAL_ROOT}/googletest)
+    endif()
+
+    if (NOT ${FETCHCONTENT_FULLY_DISCONNECTED})
```
skeetsaz, line 33: I want to be able to point it at a local GoogleTest installation instead of downloading it from Git.  Looks like this is set up so it's possible, that right?

thirtytwobits: I could wire in a redirect but you can fetch googletest once then check it all into a branch to avoid doing that elsewhere. In fact, I'm planning on releasing CETL with all of these external dependencies already fetched so working from CETL releases you wouldn't need to be online.

include/cetl/variable_length_array.hpp
```diff
@@ -0,0 +1,2065 @@
...
+        }
+    }
+
+    //
+    // Rule of Five.
+    //
+    VariableLengthArray(const VariableLengthArray& rhs)
+        : Base(rhs)
+    {
+        Base::reserve(rhs.size(), rhs.max_size());
```
skeetsaz, line 959: Copy construction should preserve the capacity.
I want the following to work.
```
std::array<std::byte, 2 * 10 * sizeof(std::uint32)> buffer{};
cetl::pf17::pmr::monotonic_buffer_resource mbr{buffer};
cetl::pf17::pmr::polymorphic_allocator<std::uint32_t> pa{&mbr};

cetl::VariableLengthArray<std::uint32_t> foo{pa, 10};
foo.reserve(10);
for (auto i = 0; i < 5; i++) { foo.push_back(i); }

cetl::VariableLengthArray<std::uint32_t> foo2{foo};
// I've already planned my buffer and know that two instances of this VLA
// will fit.  I don't want this next statement to allocate.  I want foo2
// to have the same capacity as foo.
foo2.push_back(5);
```

include/cetl/variable_length_array.hpp
```diff
@@ -0,0 +1,2065 @@
...
+    using reference = typename std::add_lvalue_reference_t<value_type>;
+
+    ///
+    /// STL-like declaration of constant-reference type.
+    ///
+    using const_reference = typename std::add_const_t<std::add_lvalue_reference_t<value_type>>;
+
+    /// Required constructor for std::uses_allocator protocol.
+    /// @param alloc Allocator to use for all allocations.
+    explicit VariableLengthArray(const allocator_type& alloc) noexcept
```
skeetsaz, line 897: Is there a reason we don't default Allloc=std::allocator<T> and alloc=Alloc() here? (like std::vector does)

include/cetl/variable_length_array.hpp
```diff
@@ -0,0 +1,2065 @@
...
+        size_ = Base::fast_copy_construct(data_, capacity_, rhs.data_, rhs.size_, alloc_);
+    }
+
+    VariableLengthArray& operator=(const VariableLengthArray& rhs)
+    {
+        Base::copy_assign_from(rhs, rhs.max_size());
+        return *this;
+    }
+
+    VariableLengthArray(VariableLengthArray&& rhs) noexcept
```
skeetsaz, line 969: std::vector has copy/move constructors with a second parameter that allows passing in a different allocator.   Don't we want that for VLA too?

include/cetl/variable_length_array.hpp
thirtytwobits: Yeah, CETL deviates from the standard where it requires system resources. heap memory provided by the system and requested via SBRK are system resources. This is why we isolate the `new_delete_resource` in the `sys/memory_resource.hpp` header.

include/cetl/variable_length_array.hpp
thirtytwobits: If we need it I can add.

skeetsaz: I'm planning on using it.  If you want to add in a later pass, that's fine too.

thirtytwobits:
Sorry for the rebase but I wanted squash all the crap I've been picking away at so I can add the few substantive changes still pending on top of a single "fix CI" change.

IMG_4333

cetlvast/suites/docs/examples/example_06_polymorphic_alloc_deleter.cpp
```diff
@@ -0,0 +1,152 @@
...
+//![example_include]
+
+#include <unordered_map>
+#include <iostream>
+#include <string.h>
+
+#include <gtest/gtest.h>
+
+
+class MyObject final
```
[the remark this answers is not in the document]
thirtytwobits: Fixed.

include/cetl/variable_length_array.hpp
thirtytwobits: Looked into this and I don't think this is the correct default behaviour so I added #30 to follow up.

include/cetl/variable_length_array.hpp
thirtytwobits: Added the copy version since that was trivial. The move version is more work so I added #31

thirtytwobits:
@pavel-kirienko ? Can you approve?

pavel-kirienko (approved):

Issue #55, opened 2023-08-09
Add assign() to VLA


thirtytwobits:
Add the assign method as defined for std::vector to the VLA

PR #56, opened 2023-08-15
#55 adding assign to VLA

thirtytwobits:


thirtytwobits:
Looks like I own SonarQube some more unit tests but this is ready for your review @skeetsaz .

cetlvast/suites/unittest/test_variable_length_array_bool.cpp
```diff
@@ -231,4 +231,112 @@ TYPED_TEST(VLABoolTests, TestBoolIterator)
...
     ASSERT_EQ(true, foo.cbegin()[2]);
+
+    typename decltype(foo)::const_iterator const_begin_iterator =
+        const_cast<typename std::add_pointer<typename std::add_const<decltype(foo)>::type>::type>(&foo)->begin();
+    typename decltype(foo)::const_iterator const_end_interator =
+        const_cast<typename std::add_pointer<typename std::add_const<decltype(foo)>::type>::type>(&foo)->end();
+
+    ASSERT_EQ(false, *const_begin_iterator);
+    ASSERT_EQ(true, *(const_end_interator - 1));
+    ASSERT_EQ(true, const_begin_iterator[2]);
```
skeetsaz, line 242: Is this correct?  foo at index 2 is false, not true.

cetlvast/suites/unittest/test_variable_length_array_bool.cpp
```diff
@@ -231,4 +231,112 @@ TYPED_TEST(VLABoolTests, TestBoolIterator)
...
+    {
+        ASSERT_EQ(true, array[i]);
+    }
+    // Remember that, if we resize down, the default argument is not used for anything.
+    array.resize(9, false);
+    ASSERT_EQ(9, array.size());
+    ASSERT_EQ(false, array[0]);
+    for (std::size_t i = 2; i < array.size(); ++i)
+    {
+        ASSERT_EQ(true, array[i - 1]);
```
skeetsaz, line 298: Seems like loop and index should match earlier loop and index (i.e. start at i=1 and array[i])

cetlvast/suites/unittest/test_variable_length_array_compat.cpp
```diff
@@ -1036,3 +1189,68 @@ TEST(VLATestsNonTrivialSpecific, TestMoveAssignment)
...
+        }
+    }
+};
+
+TYPED_TEST(VLATestsGeneric, TestResizeExceptionFromCtorOnResize)
+{
+    std::allocator<Grenade>                                     allocator{};
+    cetl::VariableLengthArray<Grenade, std::allocator<Grenade>> subject{{Grenade{1}}, allocator};
+    ASSERT_EQ(1, subject.size());
+    ASSERT_THROW(subject.resize(2, Grenade{2}), GrenadeError);
```
skeetsaz, line 1253: This throws before the call to resize() doesn't it?  So what is it testing?

include/cetl/variable_length_array.hpp
```diff
@@ -1226,6 +1226,54 @@ class VariableLengthArray : protected VariableLengthArrayBase<T, Allocator>
...
+        CETL_DEBUG_ASSERT(size() > 0, "CDE_vla_005: Calling front() on an empty array is undefined.");
+        return data_[0];
+    }
+
+    /// Returns a reference to the last element in the array.
+    /// Calling this method on an empty array is undefined.
+    /// @return Reference to the last element in the array.
+    constexpr reference back()
+    {
+        // a, perhaps naive, attempt to constrain undefined behaviour considering
```
skeetsaz, line 1252: Why would you want to constrain undefined behavior?

cetlvast/suites/unittest/test_variable_length_array_bool.cpp
thirtytwobits: This is a bool array. There is no 2 unless you are running on an unbalanced ternary machine. Unfortunately CETL does not support trit boolean values. Sorry.

cetlvast/suites/unittest/test_variable_length_array_compat.cpp
thirtytwobits: image

cetlvast/suites/unittest/test_variable_length_array_bool.cpp
thirtytwobits: Good call. It's correct but wonky. Will fix.

cetlvast/suites/unittest/test_variable_length_array_bool.cpp
skeetsaz: Oh I missed that on line 193 the iterator is on index 2 and it sets the value to true.  Nevermind.

include/cetl/variable_length_array.hpp
thirtytwobits: It's a crap shoot to be sure but, if you are debugging a misuse of this method it'll be easier to find the problem in your disassembly if the segfault was really close to this method instead of off in the greater weeds. Frankly, this isn't a tested property but it's all undefined behaviour so I could just as well have pointed the reference to a full-text copy of James Joyce's Ulysses and I wouldn't be violating the interface contract (why does CETL add so much bloat to my firmware!? It's like they embedded War and Peace in 4-byte unicode into their library...)

Issue #66, opened 2023-10-06
std::make_reverse_iterator build break with VLA<bool> cbegin/cend


skeetsaz:
Trying to use std::make_reverse_iterator with cbegin/cend does not build.
```
// Added to cetlvast/suites/unittest/test_variable_length_array_bool.cpp
TYPED_TEST(VLABoolTests, TestMakeReverseConstIterator)
{
    auto array = TypeParam::make_bool_container();
    array.assign(0, true);
    array.assign(1, false);
    array.assign(2, false);
    auto rbegin = std::make_reverse_iterator(array.cend());
    auto rend = std::make_reverse_iterator(array.cbegin());
    auto c = 0;
    for(auto i = rbegin; i < rend; ++i)
    {
        if (c == 0)
        {
            ASSERT_FALSE(*i);
        }
        else if (c == 1)
        {
            ASSERT_FALSE(*i);
        }
        else {
            ASSERT_TRUE(*i);
        }
        ++c;
    }
}
```
Error:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_iterator.h:244:9: error: no viable conversion from returned value of type 'decltype(std::declval<const cetl::VariableLengthArray<bool, cetl::pf17::pmr::polymorphic_allocator<unsigned char>>>()[0])' (aka 'bool') to function return type 'std::reverse_iterator<cetl::VariableLengthArray<bool, cetl::pf17::pmr::polymorphic_allocator<unsigned char>>::IteratorImpl<const cetl::VariableLengthArray<bool, cetl::pf17::pmr::polymorphic_allocator<unsigned char>>>>::reference' (aka 'cetl::VariableLengthArray<bool, cetl::pf17::pmr::polymorphic_allocator<unsigned char>>::reference')
        return *--__tmp;
               ^~~~~~~~
```

PR #73, opened 2023-10-23
Fix for #66

thirtytwobits:
STL iterator expects reference type to be const based on if it is a const iterator or not.

### DIFF
--- a/cetlvast/suites/unittest/test_variable_length_array_compat.cpp
+++ b/cetlvast/suites/unittest/test_variable_length_array_compat.cpp
@@ -618,3 +618,59 @@ TYPED_TEST(VLATestsCompatAnyType, TestInitFromString)
     ASSERT_EQ('s', subject[0]);
     ASSERT_EQ('.', subject[subject.size() - 1]);
 }
+
+TYPED_TEST(VLATestsCompatAnyType, TestMakeReverseIterator)
+{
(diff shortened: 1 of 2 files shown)


cetlvast/suites/unittest/test_variable_length_array_compat.cpp
```diff
@@ -618,3 +618,59 @@ TYPED_TEST(VLATestsCompatAnyType, TestInitFromString)
...
+}
+
+TYPED_TEST(VLATestsCompatAnyType, TestMakeReverseConstIterator)
+{
+    std::allocator<bool> allocator{};
+    using SubjectType = typename TestFixture::template TestSubjectType<bool, std::allocator<bool>>;
+    SubjectType subject{allocator};
+    subject.assign(0, true);
+    subject.assign(1, false);
+    subject.assign(2, false);
```
skeetsaz, line 657: I am confused by the multiple calls to assign() in these tests.  That's https://github.com/OpenCyphal/CETL/blob/910aed3a85847e0e39d8d81a2a048cf8fb394cf4/include/cetl/variable_length_array.hpp#L2366, right? setting the contents to count values of T?

thirtytwobits: erp...

skeetsaz (approved):

PR #76, opened 2023-10-27
Fix issues with bool VariableLengthArray (issues 74 and 75)

skeetsaz:
Fix iterator range constructor signatures to be consistent with type T constructors.
Consistently treat size and max size in terms of number of bits.

### DIFF
--- a/include/cetl/variable_length_array.hpp
+++ b/include/cetl/variable_length_array.hpp
@@ -2121,10 +2121,6 @@ public:
     /// is out of memory this method will still return the maximum number of elements that could
     /// be stored if the allocator had enough memory, however, it will always return the maximum
     /// size passed into the constructor if that value is less than the allocator's max_size.
-    /// Note that the underlying size_ and capacity_ of VariableLengthArrayBase are in terms
-    /// of bytes and then converted to bits in size() and capacity().  However the max_size_max_
-    /// is in terms of bits and thus no conversion is needed in max_size().
-    ///
     /// @return The maximum number of elements that could be stored in this container.
     constexpr size_type max_size() const noexcept
     {
@@ -2136,7 +2132,7 @@ public:
         const size_type max_alloc_bytes =
             std::min(max_diff_bytes, std::allocator_traits<allocator_type>::max_size(alloc_));
         const size_type max_bytes = std::min(max_alloc_bytes, numeric_limits_max_bits() / 8U);
-        return std::min(max_size_max_, max_bytes * 8U);
+        return std::min(max_size_max_bits(), max_bytes * 8U);
     }
 
     /// Reduce the amount of memory held by this object to the minimum required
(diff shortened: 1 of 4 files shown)


include/cetl/variable_length_array.hpp
```diff
@@ -2404,9 +2420,26 @@ class VariableLengthArray<bool, Allocator> : protected VariableLengthArrayBase<u
...
+            clamped_capacity = max_size;
+#endif
+        }
+        Base::reserve(bits2bytes(clamped_capacity), bits2bytes(max_size));
+    }
+
     constexpr static size_type round8(const size_type value) noexcept
     {
-        return (value + 7U) & ~7U;
+        static_assert(numeric_limits_max_bits() + 7U > numeric_limits_max_bits(), "Uh oh, overflow");
```
thirtytwobits, line 2441: why is this static_assert here and not in numeric_limits_max_bits()?

include/cetl/variable_length_array.hpp
```diff
@@ -2126,14 +2121,22 @@ class VariableLengthArray<bool, Allocator> : protected VariableLengthArrayBase<u
     /// is out of memory this method will still return the maximum number of elements that could
     /// be stored if the allocator had enough memory, however, it will always return the maximum
     /// size passed into the constructor if that value is less than the allocator's max_size.
+    /// Note that the underlying size_ and capacity_ of VariableLengthArrayBase are in terms
+    /// of bytes and then converted to bits in size() and capacity().  However the max_size_max_
```
thirtytwobits, line 2125: please rename `max_size_max_` to `max_size_max_bits_`

include/cetl/variable_length_array.hpp
```diff
@@ -2404,9 +2420,26 @@ class VariableLengthArray<bool, Allocator> : protected VariableLengthArrayBase<u
...
+#endif
+        }
+        Base::reserve(bits2bytes(clamped_capacity), bits2bytes(max_size));
+    }
+
     constexpr static size_type round8(const size_type value) noexcept
     {
-        return (value + 7U) & ~7U;
+        static_assert(numeric_limits_max_bits() + 7U > numeric_limits_max_bits(), "Uh oh, overflow");
+        return (value + 7U) & ~7ULL; // Careful, "ULL" to make sure the mask is long enough
```
thirtytwobits, line 2442: should we do `~static_cast<size_type>(7U)` instead of `~7ULL`?

thirtytwobits (changes requested):

include/cetl/variable_length_array.hpp
skeetsaz: I put it here because this is where the +7U operation is performed so it seemed like it should go with that.  I could move it to numeric_limits_max_bits() if you want.

include/cetl/variable_length_array.hpp
skeetsaz: It's in the base class.  However it is undesirable for it to have different semantic meanings in the subclasses though.  I'll make a max_size_max_bits() function similar to size_bits() and capacity_bits().  That should make it slightly clearer.

include/cetl/variable_length_array.hpp
skeetsaz: Yes, that is better

thirtytwobits (approved):

Issue #77, opened 2023-11-03
Valgrind error "Mismatched free() / delete / delete []" with MaxAlignNewDeleteResource


skeetsaz:
MaxAlignNewDeleteResource mixes std::realloc with operator new and operator delete, causing Valgrind memcheck to warn "Mismatched free() / delete / delete []".

PR #78, opened 2023-11-03
Use std::malloc/std::free to be consistent with std::realloc (issue 77)

skeetsaz:
https://github.com/OpenCyphal/CETL/issues/77


thirtytwobits (approved):
Thanks for the contributions citizen!

Issue #80, opened 2023-11-10
VLA resize() does not respect max_size_max


skeetsaz:
You can resize the container to beyond its max_size_max.

PR #81, opened 2023-11-10
Fix resize() so that it respects max_size_max

skeetsaz:
https://github.com/OpenCyphal/CETL/issues/80

### DIFF
--- a/cetlvast/suites/unittest/test_variable_length_array_bool.cpp
+++ b/cetlvast/suites/unittest/test_variable_length_array_bool.cpp
@@ -338,7 +338,7 @@ TYPED_TEST(VLABoolTests, TestBoolResizeOneBit)
 TYPED_TEST(VLABoolTestsVLAOnly, TestBoolExceedingMaxSizeMax)
 {
     std::size_t max_size_max = 1ul;
-    auto array = TypeParam::make_bool_container(1ul);
+    auto array = TypeParam::make_bool_container(max_size_max);
     array.push_back(true);
     ASSERT_EQ(1, array.size());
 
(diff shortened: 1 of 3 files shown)


include/cetl/variable_length_array.hpp
```diff
@@ -1424,18 +1425,10 @@ class VariableLengthArray : protected VariableLengthArrayBase<T, Allocator>
...
             return;
         }
 
-        if (nullptr == push_back_impl(std::move(value)))
-        {
-#if defined(__cpp_exceptions)
-            throw std::length_error("size is at capacity. Use reserve to grow the capacity.");
-#endif
-        }
+        push_back_impl(std::move(value));
```
thirtytwobits, line 1431: `std::forward`

cetlvast/suites/unittest/test_variable_length_array_bool.cpp
```diff
@@ -335,6 +335,38 @@ TYPED_TEST(VLABoolTests, TestBoolResizeOneBit)
     ASSERT_EQ(0, array[4]);
 }
 
+TYPED_TEST(VLABoolTestsVLAOnly, TestBoolExceedingMaxSizeMax)
```
thirtytwobits, line 338: Can this be part of `VLABoolTests`? If not then what's the difference between vector and VLA here?

skeetsaz: No.  This test makes use of max_size_max which std::vector does not have.

include/cetl/variable_length_array.hpp
skeetsaz: Fixed

thirtytwobits (approved):

Issue #51, opened 2023-07-21
Add Polymorphic Type system


thirtytwobits:
Don't call it a COM-back, it's been here for years...

Non-reference counting, COM-like interface and associated utilities for explicit runtime-type information and safe polymorphism without using generics. This is useful for high-criticality systems where visibility of all source code is imperative and where C++ RTTI does not meet this source-visibility requirement.

```c++
/// 128-bit type ID for polymorphic types. On most platforms, this will be represented as a 16-byte array but some
/// platforms may end up optimizing this to use 128-bit SIMD instructions. As such, USE THE TYPE ALIAS!
using PolymorphicTypeId = std::array<std::uint8_t, 16>;

/// Base class for polymorphic types that support a runtime type system. This is used instead of RTTI as it is fully
/// visible in the source code and applies only to a specific set of types.
class IPolymorphicType
{
public:
    /// Query the type to see if it supports a given type. Mutable override.
    /// @param id   The UID of the type to query. This ID is unique to a single libcyphal polymorphic type.
    /// @param out  A pointer that will be set to an interface supporting IPolymorphicType but which can be safely cast
    ///             to a known subtype. This is set to nullptr for all failures.
    ///             @note Libcyphal does not use object composition. As such, the pointer returned will always be to a
    ///             subtype of the object being queried.
    /// @return true if the type is supported and the out pointer was set. false otherwise.
    virtual bool queryType(const PolymorphicTypeId& id, void*& out) noexcept = 0;

    /// Query the type to see if it supports a given type. Const override.
    /// @param id   The UID of the type to query. This ID is unique to a single libcyphal polymorphic type.
    /// @param out  A pointer that will be set to an interface supporting IPolymorphicType but which can be safely cast
    ///             to a known subtype. This is set to nullptr for all failures.
    ///             @note Libcyphal does not use object composition. As such, the pointer returned will always be to a
    ///             subtype of the object being queried.
    /// @return true if the type is supported and the out pointer was set. false otherwise.
    virtual Status queryType(const PolymorphicTypeId& id, const void*& out) const noexcept = 0;

    /// Test if two instances of the type are the same instance. Polymorphic types are free to define instance equality
    /// as they see fit.
    /// @param other The other instance to compare to.
    /// @return true if this instance is equal to other else false.
    virtual bool isEqual(const IPolymorphicType& other) const noexcept = 0;

protected:
    /// Polymorphic types cannot be deleted, copied, or moved through this interface.
[...]

Issue #85, opened 2024-01-10
Question about the naming convention


pavel-kirienko:
If CETL is intended to resemble `std`, should it not use the same naming convention? One significant deviation is that it uses camelCase or PascalCase where std uses snake_case. There already are some entities that would need to be renamed, but it can be done with (deprecated) aliases to avoid API breakage.

pavel-kirienko:
We are already accepting PRs with the new naming convention (#89, #90, #92) so this is considered decided.

PR #89, opened 2024-02-09
Polyfill cetl::optional, extend the testing suite

pavel-kirienko:
- Add `optional` and related entities such as `make_optional`, `nullopt`, `in_place`, etc.
- Add `type_traits.hpp`.
- Fix usage of `CETL_PROJECT_VERSION` in the build recipes.
- Disable C++ extensions in CETLVaSt (they were mistakenly left enabled).
- Add basic typelist operations to CETLVaSt that are needed for exhaustive combinatorial testing.
- Add the `CETL_NODISCARD` attribute.
- Add a useful feature to the GitHub Actions workflow: run CI jobs for all pushes where the head commit contains a hashtag with the name of the job; e.g., adding `#verification` to the commit message will run the verification job.

This PR is feature- and testing-complete but lacks the docs, which I am going to add next. The code is ready to be reviewed meanwhile. You may notice that the coverage reporting is odd because there are a few places where templates are not marked as instantiated while they actually are, also some branches in the comparison operators appear to be mislabeled as partially-covered.

The commit history is non-informative and should not be looked at, as it will be squashed anyway.

My next step is to finish #51 (which is blocked on `cetl::optional`) and then perhaps switch to #86, because there are certain similarities between `optional` and `variant` that will make it easier for me to do this part now rather than later.


pavel-kirienko:
Regarding the naming convention used see https://github.com/OpenCyphal/CETL/issues/85

thirtytwobits (changes requested):
Still reviewing but I'll flush these comments. Most important is `is_nothrow_swapable` is not in C++14.

include/cetl/pf17/optional.hpp
```diff
@@ -0,0 +1,883 @@
...
+    {
+        reset();
+        new (std::addressof(this->m_value)) T(il, std::forward<Args>(args)...);
+        this->m_engaged = true;
+        return this->m_value;
+    }
+
+    /// Swaps two optionals. If either is not engaged, acts like move assignment.
+    void swap(optional& other) noexcept(std::is_nothrow_move_constructible<T>::value&&  //
+                                            std::is_nothrow_swappable<T>::value)
```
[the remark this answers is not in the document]
pavel-kirienko: Yeah, this is because we had C++ extensions enabled. Both GCC and Clang expose `std::is_nothrow_swappable` in `-std=gnu++14` mode but not in the strict `-std=c++14` mode; see <https://godbolt.org/z/h9PjeE7sG>. Our CMake recipe had a statement `set(CXX_EXTENSIONS OFF)` but the correct variable for this is `CMAKE_CXX_EXTENSIONS` unless you set it as a per-target property.

pavel-kirienko:
Heads up: I tweaked two Doxygen settings to eliminate unhelpful warnings about undocumented enum members (well-named entities may not require explicit docs) and to automatically distribute docs to all group members to reduce doc copypasta.

lydia-at-amazon (approved):
Perused the code. Looks good. Should definitely be enough to fulfill our immediate use case in Libcyphal. Approving

cetlvast/include/cetlvast/typelist.hpp
```diff
@@ -0,0 +1,156 @@
...
+template <typename A, typename B, typename... C>
+struct cat<A, B, C...>
+{
+    using type = typename cat<typename cat<A, B>::type, C...>::type;
+};
+}  // namespace _impl_cat
+
+// ------------------------------------------------------------------------------------------------
+
+/// TYPELIST CONCATENATOR: (list(A...), list(B...), list...) -> list(A..., B..., ...)
```
thirtytwobits, line 40: nifty

include/cetl/pf17/attribute.hpp
```diff
@@ -0,0 +1,26 @@
...
+/// Copyright Amazon.com Inc. or its affiliates.
+/// SPDX-License-Identifier: MIT
+
+#ifndef CETL_PF17_ATTRIBUTE_HPP_INCLUDED
+#define CETL_PF17_ATTRIBUTE_HPP_INCLUDED
+
+/// CETL_NODISCARD
+#ifndef CETL_NODISCARD
+#    if (__cplusplus >= 201703L) || defined(CETL_DOXYGEN)
+/// A compatibility macros that expands to \c [[nodiscard]] if C++17 or later is used, otherwise it expands to
```
thirtytwobits, line 14: I think you need to start this with `@def` for doxygen to pick it up.

include/cetl/_helpers.hpp
```diff
@@ -0,0 +1,94 @@
+/// @copyright
```
thirtytwobits, line 1: I want to avoid widely entangled dependencies in CETL. Can we rename this to `_helper_enable_copy_move_construction.hpp` to avoid future abuse of this header as a dumping ground for syntactic sugar?

include/cetl/pf17/attribute.hpp
thirtytwobits: Ah. Okay.

thirtytwobits (approved):
Please add a task to create a `cetlvast/suites/docs/examples/` example for optional in the future.

PR #90, opened 2024-02-19
Add polymorphic type system

pavel-kirienko:
This is a work in progress: the design may require another pass, the docs are not pretty, and usage examples are missing.

Closes https://github.com/OpenCyphal/CETL/issues/51

Regarding the naming convention used, see https://github.com/OpenCyphal/CETL/issues/85

### Questions

- ~~Do we allow an exception to the no-macro policy for the `CETL_RTTI` macro currently defined in the unit test file? I would like to promote it from the test suite to the library API. I considered CRTP as an alternative like `class your_class : cetl::rtti_impl<your_class>`, but this approach runs into the diamond problem and the bases of `your_class` are not easily accessible from the CRTP class.~~
  - ~~Pros: reduces boilerplate and minimizes the risk of copy-paste errors.~~
  - ~~Cons: it's a macro.~~

- EDIT: Disregard the question about the macro, I found a different solution based on a helper class; see the usage example below.

- Do the semantics make sense?

- Do we want to bikeshed the naming a bit?

- I couldn't find a use case for `isEqual` proposed in #51; am I missing something?

- Can I allow single-line methods in the Clang-Format config?

### Usage example

Something similar should be provided in the docs.

```c++
#include <cetl/rtti.hpp>

// A user-defined type that has opted into the RTTI support by inheriting from rtti_helper.
struct IfaceA : cetl::rtti_helper<cetl::type_id_type<0xde, 0xad>>
{
    // ...
};

// An unrelated type that is also RTTI-capable.
struct IfaceB : cetl::rtti_helper<cetl::type_id_type<0xbe, 0xef>>
{
    // ...
};

// Some class that implements both interfaces.
// The quirk of this design is that we inherit from the interface classes
// not directly but via rtti_helper; this is needed to inform the helper
// about the types in the hierarchy.
struct Impl : cetl::rtti_helper<cetl::type_id_type<0xca, 0xca>, IfaceA, IfaceB>
{
    // Implement IfaceA and IfaceB.
};
```

An attentive reader will notice that we have a diamond inheritance on our hands here because:

```mermaid
classDiagram
    `cetl::rtti` <|-- IfaceA
    `cetl::rtti` <|-- IfaceB
    IfaceA <|-- Impl
    IfaceB <|-- Impl
```

This is still compliant with AUTOSAR A10-1-1 by exception because all base classes involved are interface classes. You can safely convert between them arbitrarily, up and down, using `cetl::rtti_cast`.

Another quirk is that up-conversions are performed through the RTTI mechanism (using `cetl::rtti::_cast_`) while a `static_cast` would suffice; this is trivial to change though if this design is accepted. The C++ standard prohibits conversions to the diamond base because of the ambiguity involved; our solution, however, silently resolves the ambiguity during up-conversions by choosing the first base (this is due to the way `cetl::rtti_helper::search` is implemented).


thirtytwobits:
> 
> An attentive reader will notice that we have a diamond inheritance on our hands here because:
> 
> ```mermaid
> classDiagram
>     `cetl::rtti` <|-- IfaceA
>     `cetl::rtti` <|-- IfaceB
>     IfaceA <|-- Impl
>     IfaceB <|-- Impl
> ```

Of course, a good compiler will elide all of this with interfaces being pure-virtual but even-so, since you are not using virtual inheritance your diagram should read:

```mermaid
classDiagram
     `cetl::rtti::0` <|-- IfaceA
     `cetl::rtti::1` <|-- IfaceB
     IfaceA <|-- Impl
     IfaceB <|-- Impl
```

Which begs the question, should we use virtual inheritance where an interface inherits another interface?

thirtytwobits:
> Another quirk is that up-conversions are performed through the RTTI mechanism (using cetl::rtti::_cast_) while a static_cast would suffice; this is trivial to change though if this design is accepted. The C++ standard prohibits conversions to the diamond base because of the ambiguity involved; our solution, however, silently resolves the ambiguity during up-conversions by choosing the first base (this is due to the way cetl::rtti_helper::search is implemented).

Using virtual inheritance between interfaces removes this ambiguity meaning we don't need the `rtti_helper::search`

include/cetl/rtti.hpp
```diff
@@ -0,0 +1,316 @@
...
+    /// Implementations are adivsed to use \ref cetl::rtti_helper instead of implementing this manually.
+    /// If manual implementation is preferred, then the method body should simply forward the result of this
+    /// type's static \c _get_static_type_id_ method:
+    ///
+    /// @code
+    /// return _get_static_type_id_();
+    /// @endcode
+    ///
+    /// This method should not be invoked directly; instead, use \ref cetl::get_type_id.
+    CETL_NODISCARD virtual type_id _get_polymorphic_type_id_() const noexcept = 0;
```
thirtytwobits, line 121: Which ID is returned though? Polymorphic objects can be multiple types. That's the whole point. By asking "are you this type" you don't have to expose a list of types to the user which is what you'll have to do if you are going to ask for types from the object.

By moving away from the "queryType" method you remove the ability for implementations to be returned directly instead of using delegated interfaces. This might be okay. It does simplify the type system even if it means certain optimizations, normally available to COM objects, are not supported and it means we don't add anything that rtti doesn't provide.

include/cetl/rtti.hpp
```diff
@@ -0,0 +1,316 @@
...
+    /// should be checked in the same way one by one and the first match (non-nullptr result) should be returned.
+    ///
+    /// The objective here is to check if the current type in the hierarchy is the one being sought;
+    /// if not, delegate the call to the base class(es); the topmost class is this \c rtti type that
+    /// simply returns a null pointer, indicating that the search has returned no matches and a
+    /// safe dynamic type conversion is not possible (at least not along the current branch of the type tree).
+    ///
+    /// This method should not be invoked directly; instead, use \ref cetl::rtti_cast.
+    /// @{
+    CETL_NODISCARD virtual void* _cast_(const type_id& id) & noexcept
```
thirtytwobits, line 142: These need to be pure-virtual. Everything in the base class has to be pure-virtual to ensure the compiler can completely elide the interface types when optimizing.

include/cetl/rtti.hpp
```diff
@@ -0,0 +1,316 @@
...
+{
+    return get_type_id<T>();
+}
+
+/// Performs a safe dynamic type up-/down-conversion in constant time by invoking \ref cetl::rtti::_cast_.
+/// \c T shall be a pointer and \c std::remove_pointer_t<T> shall satisfy \ref cetl::has_static_type_id.
+/// Returns \c nullptr if a safe dynamic type conversion to \c T is not possible.
+/// @{
+template <typename T, typename _from>
+CETL_NODISCARD std::enable_if_t<is_rtti_convertible<_from> &&     //
```
thirtytwobits, line 270: Doesn't this require the caller to have access to the concrete type? That sorta defeats the whole purpose.

thirtytwobits (changes requested):
I don't think this is correct. Specifically, I'm not convinced you can do a safe down-cast using this implementation unless you have access to the concrete type.

include/cetl/rtti.hpp
pavel-kirienko: >Which ID is returned though? Polymorphic objects can be multiple types.

The ID of the most concrete type, aka runtime type, is returned, which is the behavior of `dynamic_cast`. I can spell it out in the docs to avoid misinterpretations.

>By moving away from the "queryType" method you remove the ability for implementations to be returned directly instead of using delegated interfaces.

I attempted to deliver the functionality offered by queryType in this design via the `_cast_` method, and its user-facing facade `rtti_cast`. Do you think this design does not offer equivalent capabilities to the user?

include/cetl/rtti.hpp
pavel-kirienko: I understand this but do note that pure virtual methods here would require special-casing in the search method when the next base to query is `cetl::rtti`. Do we need to sacrifice generality for the sake of optimization?

include/cetl/rtti.hpp
pavel-kirienko: No. It only requires the caller to have access to the target type for the conversion, which may be different from the runtime type (concrete type).

include/cetl/rtti.hpp
thirtytwobits: It's not an optimization. It's a standard saying that we do not utilize any concrete, multiple-inheritance. Adding these methods breaks that promise. It may also expose us to edge cases with destructors and other nasties. Rather than try to reason about all of these we should just not have concrete implementation in interface classes.

Also, you don't need the search if you use virtual inheritance, right?

include/cetl/rtti.hpp
thirtytwobits: My spidey senses are tingling here. I'm not sure it's safe to allow the caller to adjust the pointer (i.e. static_cast) in every circumstance. The queryType method (which is based on techniques that are at least 26 years old) encapsulates any pointer adjustments so the caller doesn't need to reason about them. The compiler will have full visibility into the types involved when it parses the queryType implementation and the only unsafe code is the reinterpret_cast of the returned void* but we know that this pointer is correct and should not be adjusted.

I also don't like the need to extend a template. The whole idea of this type system is that any template usage is pure syntactic sugar. My Janky proposal did not use templates to be a polymorphic type which kept things more POCO-like. The `polymorphic_type_traits` was purely a convenience. The types could be used without any templates if needed.

I'm open to improvements, be they simplifications or worthwhile enhancements, to my IPolymorphicType proposal but I think this current proposal is inferior.

include/cetl/rtti.hpp
pavel-kirienko: Ok, I understand your point, but the search is not related to the kind of inheritance used. It is needed to traverse the tree of inheritance to find the requested type in it, starting from the most concrete type (the runtime type) and then climbing upwards each branch of the tree.

include/cetl/rtti.hpp
pavel-kirienko: Sorry, I still don't understand the objection.

>The compiler will have full visibility into the types involved when it parses the queryType implementation and the only unsafe code is the reinterpret_cast of the returned void* but we know that this pointer is correct and should not be adjusted.

Maybe I'm missing something but this is precisely how my implementation works.

1. The client calls `_cast_` (not directly in my implementation but via a convenience helper, but it's optional).
2. The `_cast_` method is executed on the actual runtime type where a check is done whether the requested type is the same as the runtime type.
  - If the type matches, `static_cast<void*>(this)` is returned. We know that `static_cast<T*>(static_cast<void*>(this))` is safe for any `T` always; the safety of the conversion is not undermined if the inner part is performed inside `_cast_` and the outer part is performed outside of it by the `rtti_cast` helper; hence the whole conversion is safe.
  - If the type does not match, `_cast_` of the first base is invoked; if it returns non-null, its result is returned.
  - Otherwise, the search continues to the next base until the first non-null result or all bases have been traversed.
3. The result is `static_cast` into the requested type. The `_cast_` method has ascertained that the conversion is safe, as shown above.

>I also don't like the need to extend a template. The whole idea of this type system is that any template usage is pure syntactic sugar. My Janky proposal did not use templates to be a polymorphic type, which kept things more POCO-like. The polymorphic_type_traits was purely a convenience. The types could be used without any templates if needed.

I'm not sure about the template either, there are pros and cons. The original design I submitted when I just opened this draft PR did not use the template helper; instead it used a macro to avoid boilerplate. You can inspect it here: https://github.com/OpenCyphal/CETL/pull/90/commits/561ce22477b43c998ef11c1e0366e8606742d68c

Here's the definition of the macro; it is meant to be expanded in the class body:

```c++
/// An optional helper that can be used to implement CETL RTTI support with minimal boilerplate.
/// Use it in the public section of the class definition.
///
/// The first argument is the base class of the class that is being defined, which can be \c cetl::rtti.
///
/// The following arguments are the 16 bytes of the type identifier exposed via \c _get_static_type_id_;
/// if less than 16 bytes are provided, the remaining bytes are zeroed.
#define CETL_RTTI(base, ...)                                                                       \
    static constexpr cetl::type_id _get_static_type_id_() noexcept                                 \
    {                                                                                              \
        return cetl::type_id{__VA_ARGS__};                                                         \
    }                                                                                              \
    cetl::type_id _get_polymorphic_type_id_() const noexcept override                              \
    {                                                                                              \
        return _get_static_type_id_();                                                             \
    }                                                                                              \
    void* _cast_(const cetl::type_id& id)& noexcept override                                       \
    {                                                                                              \
        return (id == _get_static_type_id_()) ? static_cast<void*>(this) : base::_cast_(id);       \
    }                                                                                              \
    const void* _cast_(const cetl::type_id& id) const& noexcept override                           \
    {                                                                                              \
        return (id == _get_static_type_id_()) ? static_cast<const void*>(this) : base::_cast_(id); \
    }
```

This is not a very large macro so one should be able to comfortably survive without it. Its downside is that it doesn't support multiple inheritance; as you can see, it takes only one base, meaning that if T implements more than one interface, it will not be able to cast to a type located in the inheritance branch of the non-first base. To support multiple inheritance, one would need to either use the `rtti_helper<>` base or to traverse the bases manually.

As far as I see, the same considerations are still applicable to your proposal, unless I'm misunderstanding something.

thirtytwobits: Macros are proscribed for anything that isn't conditional includes. CETL takes minimal exceptions for this when not using the polyfill headers (e.g `#if !defined(__cpp_exceptions)`). The polyfill type system must not require the use of macros or all objects using it will require exceptions to the no-macros rule.

The whole reason for this polymorphic type system is to avoid generic programming where this can be considered problematic. The design we come up with here must work without macros nor templates but can provide template helpers. `IPolymorphicType` was a simple C++ interface. We would be able to define and implement polymorphic objects without macros or templates. Use of these is trivial without the template helpers I provided making them pure syntactic sugar. For example:

[Example without any helpers...](https://godbolt.org/z/v6zseEaGn)

[Same example with helpers...](https://godbolt.org/z/YoYcb5asd)

As you can see the helpers only serve to eliminate `reinterpret_cast` from user code and the design is completely useable without them.

thirtytwobits: Finally, `queryType` Allows for true object composition without delegation:

[composition example...](https://godbolt.org/z/baW8jffos)

pavel-kirienko: Okay, we seem to be on the same page about the behavioral part of the implementation, as mine does precisely the same thing as yours but with a slightly different API. The composition is also possible as you described:

```c++
class CatDog final : public virtual cetl::rtti
{
  public:
    void* _cast_(const type_id& id) & noexcept override
    {
        if (void* const p = m_cat._cast_(id)) {
            return p;
        }
        return m_dog._cast_(id);
    }

    const void* _cast_(const type_id& id) const& noexcept override
    {
        if (const void* const p = m_cat._cast_(id)) {
            return p;
        }
        return m_dog._cast_(id);
    }

  private:
    Tabby m_cat{};
    Boxer m_dog{};
};
```

The difference is that I put a much stronger emphasis on the use of helpers than you did. I want you to notice that an equally generic implementation of your approach will also require the search method; what you have in the CatDog example is a very special-case implementation of the same search method.

I have identified the following deficiencies in my proposal so far:

- [ ] Use virtual inheritance instead of ordinary inheritance to avoid ambiguities during up-conversions (see the OP post).

- [ ] Update the docs to de-emphasize the use of helpers in favor of manual implementation. Keep the `cetl::rtti_helper<>` in place for the benefit of less template-conscious applications and as a usage example.

- [ ] Make all methods of `cetl::rtti` pure. The `cetl::rtti_helper` will require specialization to properly terminate the search when the top of a branch is reached.

- [ ] Upon a closer look I see that `_get_polymorphic_type_id_` is inessential and can be removed; `_cast_` is sufficient for all uses, including type identity queries (see `cetl::is_instance_of`).

include/cetl/rtti.hpp
```diff
@@ -0,0 +1,297 @@
+/// @file
+/// A simple runtime type information (RTTI) support designed for high-integrity
+/// real-time systems, where the use of the standard C++ RTTI is discouraged.
```
thirtytwobits, line 3: How about this for the header comment:

> An explicit mechanism for implementing dynamic type introspection for high-integrity systems. This should be used instead of standard C++ RunTime Type Information (RTTI)  (i.e. `-f-nortti`) where a codebase must trace from lines of source back to verifications (correctness) and requirements (suitability).

thirtytwobits (approved):
One change to the header comment then ship-it. Nice work.

Issue #97, opened 2024-03-12
"invalid escape sequence `\|`" warning when running `verify.py -h`.


serges147:
On latest python 3.12.2 (macOS 14.4) when running `verify.py -h` there is following warning in the output:
```
./build-tools/bin/verify.py -h
/Users/sergei/Develop/git/OpenCyphal/CETL/./build-tools/bin/verify.py:164: SyntaxWarning: invalid escape sequence '\|'
  """
...
```

It seems that "Cyphal" logo text "image" is missing "r" (raw string) prefix (like `r"""...`).
I'm gonna fix this as part of my onboarding (fork, pr, etc.)

pavel-kirienko:
Fixed in https://github.com/OpenCyphal/CETL/pull/98

PR #98, opened 2024-03-12
Fix for #97: "invalid escape sequence \|" warning

serges147:
Making "opencyphal" text logo as a **raw** (unescaped) python string.


pavel-kirienko (approved):
:tada:

Issue #83, opened 2024-01-09
Provide PMR-aware cetl::any


pavel-kirienko:
This is like `std::any` with the following differences:

- Use PMR instead of the default heap.
- Provide configurable-size small object optimization, preferably with an option to disable PMR completely, such that if an attempt to allocate an object that does not fit into the small-object arena is made, a compile-time error will result.

pavel-kirienko:
@thirtytwobits should we at first implement small-object version only, without PMR support, similar to #84? We can add PMR support later without breaking the API, I think.

pavel-kirienko:
Here's a relevant example: [unique_any_poc.tar.gz](https://github.com/OpenCyphal/CETL/files/14556987/unique_any_poc.tar.gz)

PR #102, opened 2024-03-18
Also first revision of `any` methods, functions and tests - without RTTI yet.

serges147:
- Implemented most of constructors (close to `std::any`).
- Added implementation of 3 `operator=`-s.
- Added first 2 overloads of `any_cast`.
- Added `make_any` overloads.
- Implemented `has_value`, `emplace`, `swap` & `reset`.
- `cetl::bad_any_cast` now alias to `cetl::pf17::bad_any_cast `.

Also:
- Appended `is_in_place_type_v` and `is_in_place_index_v` constexpr-s.
- Applied clang format to `utility.hpp` header file.

Related to #83


include/cetl/pf17/any.hpp
```diff
@@ -0,0 +1,46 @@
...
+namespace cetl  // NOLINT(*-concat-nested-namespaces)
+{
+namespace pf17
+{
+
+#if defined(__cpp_exceptions) || defined(CETL_DOXYGEN)
+
+/// A polyfill for `std::bad_any_cast`.
+/// This is only available if exceptions are enabled (`__cpp_exceptions` is defined).
+class bad_any_cast : public std::bad_cast
```
pavel-kirienko, line 25: It makes sense that this class is here, as it polyfills `std::bad_any_cast`, but I wonder if it's going to be a problem later on when we need to add a specialization of our `cetl::any<>` into this file, because that would require this header to depend on `cetl/any.hpp`, which creates a recursive dependency.

If this is a valid concern, we might be better off moving everything that is needed for `cetl/any.hpp` into its own header.

include/cetl/any.hpp
```diff
@@ -0,0 +1,345 @@
...
+    {
+        static_assert(sizeof(Tp) <= Footprint, "Enlarge the footprint");
+
+        static void* handle(detail::action action, const any* self, any* other)
+        {
+            switch (action)
+            {
+            case detail::action::Get:
+
+                return get(const_cast<any&>(*self));
```
pavel-kirienko, line 218: `const_cast` violates AUTOSAR A5-2-3

include/cetl/any.hpp
```diff
@@ -0,0 +1,345 @@
...
+    base_copy& operator=(const base_copy&) = delete;
+};
+
+enum class action
+{
+    Get,
+    Copy,
+    Move,
+    Destroy
+};
```
pavel-kirienko, line 90: This code is fine but really cool kids from the block use algebraic types for this these days, because it allows one to express more concepts through the type system:

```c++
struct action_get{};
struct action_copy{ any* other; };  // notice the `other`!
struct action_move{ any* other; };  // it is introduced only when needed
struct action_destroy{};
using action = variant<action_get, action_copy, action_move, action_destroy>;
```

Then you use `visit` to handle `action` in a type-safe way without the need to introduce unused parameters.

Just saying.

include/cetl/any.hpp
```diff
@@ -0,0 +1,345 @@
...
+    // Type-erased handler.
+    using any_handler = void* (*) (detail::action, const any* /*self*/, any* /*other*/);
+
+    // Small Object Optimization (SOO) handler.
+    template <typename Tp>
+    struct soo_handler
+    {
+        static_assert(sizeof(Tp) <= Footprint, "Enlarge the footprint");
+
+        static void* handle(detail::action action, const any* self, any* other)
```
pavel-kirienko, line 212: The way I read it, non-const function parameters that are not mutated violate AUTOSAR A7-1-1.

include/cetl/any.hpp
```diff
@@ -0,0 +1,345 @@
...
+        {
+            switch (action)
+            {
+            case detail::action::Get:
+
+                return get(const_cast<any&>(*self));
+
+            case detail::action::Copy:
+
+                copy(*other, *self);
```
pavel-kirienko, line 222: I mentioned this yesterday in the evening but just for historical purposes I want to mention it here again that this code will break if `Tp` is non-copyable. Same goes for moving. This is why we need these telescoping bases here that select the copy/move behaviors depending on the template parameters.

include/cetl/any.hpp
```diff
@@ -0,0 +1,345 @@
...
+    any_handler handler_ = nullptr;
+
+};  // class any
+
+/// \brief Constructs an any object containing an object of type T, passing the provided arguments to T's constructor.
+///
+/// Equivalent to `cetl::any(cetl::in_place_type<ValueType>, std::forward<Args>(args)...)`.
+///
+template <typename ValueType, typename Any, typename... Args>
+inline Any make_any(Args&&... args)
```
pavel-kirienko, line 296: Template functions are always `inline` by default.

Nitpick but I think we should use `CETL_NODISCARD` here and in similar cases, where not using the return value is clearly an error. It also helps enforce AUTOSAR A0-1-2.

cetlvast/suites/unittest/test_any.cpp
```diff
@@ -0,0 +1,303 @@
...
+    EXPECT_FALSE((any<13, true, false>{}.has_value()));
+}
+
+TEST(test_any, ctor_2_copy)
+{
+    // Primitive `int`
+    {
+        using uut = any<sizeof(int)>;
+
+        const uut srcAny{42};
```
pavel-kirienko, line 63: Applies in other places as well.

```suggestion
        const uut src_any{42};
```

include/cetl/pf17/any.hpp
serges147: Yes, I though about it, but decided to tackle this when I start working on the specialization. Is it fine to postpone?

include/cetl/any.hpp
serges147: I'll think about it. Also, maybe when I start working on Copyable/Movable constraints then it's possible that enum and switch won't be applicable anymore...

include/cetl/any.hpp
serges147: will fix it in this pr...

cetlvast/suites/unittest/test_any.cpp
serges147: @pavel-kirienko is it fine if I fix this in separate pr? it will be just "cosmetic" pr without functional changes.

include/cetl/any.hpp
serges147: I will eliminate all these `const_cast`-s when reworking the `handle` thing (and implementing Copyable/Movable constraints). is it ok to postpone to near future pr?

include/cetl/pf17/any.hpp
pavel-kirienko: okay

cetlvast/suites/unittest/test_any.cpp
pavel-kirienko: Why can't we fix them now?

serges147: yes we can - I just though it will be easier review separately "cosmetics" ONLY changes

pavel-kirienko: I don't think it makes a lot of difference here

serges147: ok, I'll do it now, thnx

serges147: done - see last commit 2ffc0051f5106b86dd7cdf5692f06c93fc311e35

include/cetl/any.hpp
```diff
@@ -0,0 +1,344 @@
...
+
+// Move policy.
+template <std::size_t Footprint, bool Copyable, bool Movable>
+struct base_move;
+
+// Copy policy.
+template <std::size_t Footprint, bool Copyable>
+struct base_copy;
+
+template <std::size_t Footprint, std::size_t Align = sizeof(std::max_align_t)>
```
pavel-kirienko, line 35: The alignment should be also configurable via the list of template parameters to any:

```c++
template <std::size_t Footprint,
          bool Copyable = true,
          bool Movable = Copyable,
          std::size_t Align = alignof(std::max_align_t)>
```

I neglected to mentioned this explicitly earlier. Also note that `alignof` should be used instead of `sizeof`.

pavel-kirienko (approved):

include/cetl/any.hpp
serges147: `Align` configurable will be ready (and tested) after next pr (already ready in my separate private branch).

PR #104, opened 2024-03-20
Almost finalize main api of `cetl::any` - now close to `std::any` features.

serges147:
- Added "value return" overloads of any_cast (#1️⃣, 2️⃣ & 3️⃣).
- Added std::initializer_list overloads for
  - ctor (#6️⃣)
  - `emplace` (#2️⃣)
  - `make_any` (#2️⃣)
- Added corresponding unit tests, plus more cases to increase coverage.

Notes for to be continued...
- `Align` param fixes will be in the next pr.
- Copyable/Movable (as well as elimination of various `const_cast`) will be in the after next pr.

### DIFF
--- a/cetlvast/suites/unittest/test_any.cpp
+++ b/cetlvast/suites/unittest/test_any.cpp
@@ -185,7 +185,7 @@ TEST(test_any, ctor_6_in_place_initializer_list)
         std::size_t size_;
         int         number_;
 
-        TestType(std::initializer_list<char> const chars, int const number)
+        TestType(const std::initializer_list<char> chars, const int number)
         {
             size_   = chars.size();
             number_ = number;
@@ -309,7 +309,7 @@ TEST(test_any, make_any_2_ilist)
         std::size_t size_;
         int         number_;
 
-        TestType(std::initializer_list<char> const chars, int const number)
+        TestType(const std::initializer_list<char> chars, const int number)
         {
             size_   = chars.size();
             number_ = number;
@@ -332,19 +332,7 @@ TEST(test_any, any_cast_cppref_example)
 
 #if defined(__cpp_exceptions)
 
-    try
-    {
-        (void) any_cast<std::string>(uut{});  // throws!
-        FAIL() << "Should have thrown an exception.";
-
-    } catch (const cetl::bad_any_cast& e)
-    {
-        SUCCEED() << "`bad_any_cast` has been caught.";
-
-    } catch (...)
-    {
-        FAIL() << "Should have thrown `bad_any_cast`.";
-    }
+    EXPECT_THROW(any_cast<std::string>(uut{}), cetl::bad_any_cast);
 
 #endif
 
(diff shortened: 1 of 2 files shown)


include/cetl/any.hpp
```diff
@@ -294,10 +316,84 @@ class any : private detail::base_move<Footprint, Copyable, Movable>
...
 {
-    return Any(cetl::in_place_type<ValueType>, std::forward<Args>(args)...);
+    return Any(in_place_type<ValueType>, std::forward<Args>(args)...);
+}
+
+/// \brief Constructs an any object containing an object of type T, passing the provided arguments to T's constructor.
+///
+/// Equivalent to `cetl::any(cetl::in_place_type<ValueType>, list, std::forward<Args>(args)...)`.
+///
+template <typename ValueType, typename Any, typename Up, typename... Args>
```
pavel-kirienko, line 326: I wonder if we could default Any to something like this:

```suggestion
template <T>
using any_like = any<sizeof(T),
                    std::is_copy_constructible<T>::value,
                    std::is_move_constructible<T>::value,
                    std::max(alignof(T), alignof(std::max_align_t))>;

template <typename ValueType, typename Any = any_like<ValueType>, typename Up, typename... Args>
```

I only thought about this now but it should have been during the previous review.

Do you think it makes sense? Otherwise we deviate from the semantics of std::make_any quite a bit as it doesn't require one to specify the type of any.

cetlvast/suites/unittest/test_any.cpp
```diff
@@ -272,13 +294,176 @@ TEST(test_any, make_any_1_cppref_example)
...
+#endif
+
+    // Advanced example
+    a1       = std::string("hello");
+    auto& ra = any_cast<std::string&>(a1);  //< reference
+    ra[1]    = 'o';
+    EXPECT_STREQ("hollo", any_cast<const std::string&>(a1).c_str());  //< const reference
+
+    auto s1 = any_cast<std::string&&>(std::move(a1));  //< rvalue reference
+    // Note: “s1” is a move-constructed std::string, “a1” is empty
```
pavel-kirienko, line 358: The quotes appear to violate AUTOSAR A2-3-1.

include/cetl/any.hpp
```diff
@@ -294,10 +316,84 @@ class any : private detail::base_move<Footprint, Copyable, Movable>
...
+/// \brief Performs type-safe access to the contained object.
+///
+/// \param operand Target any object.
+/// \return Returns `std::static_cast<ValueType>(*cetl::any_cast<const U>(&operand))`,
+///     where let `U` be `std::remove_cv_t<std::remove_reference_t<ValueType>>`.
+///
+template <typename ValueType, typename Any>
+ValueType any_cast(const Any& operand)
+{
+    using RawValueType = typename std::remove_cv<typename std::remove_reference<ValueType>::type>::type;
```
pavel-kirienko, line 341: It is best to use `_t`/`_v` helpers where available.

```suggestion
    using RawValueType = std::remove_cv_t<std::remove_reference_t<ValueType>>;
```

cetlvast/suites/unittest/test_any.cpp
```diff
@@ -355,7 +343,7 @@ TEST(test_any, any_cast_cppref_example)
     EXPECT_STREQ("hollo", any_cast<const std::string&>(a1).c_str());  //< const reference
 
     auto s1 = any_cast<std::string&&>(std::move(a1));  //< rvalue reference
-    // Note: “s1” is a move-constructed std::string, “a1” is empty
+    // Note: `s1` is a move-constructed std::string, `a1` is empty
```
serges147, line 346: We agreed to add back ticks "`" as an exception (b/c strictly speaking they are not allowed as well).

pavel-kirienko (approved):
assert is still missing, please see the previous review comment

PR #107, opened 2024-03-22
Reworked handling of get, copy, move and destroy functionality.

serges147:
- Now `Copyable` and `Movable` constraints are respected.
- Now instead of single handler we have 3 (destroyer, copier, mover) - allows to eliminate AUTOSAR `const_cast` violations.
- In unit tests:
  - more tests to cover combinations of `any` template params
  - switch from pointer based `any_cast` to value or reference based.

### DIFF
--- a/include/cetl/any.hpp
+++ b/include/cetl/any.hpp
@@ -300,7 +300,7 @@ public:
 
     any& operator=(const any& rhs)
     {
-        any(rhs).swap(std::move(*this));
+        any(rhs).swap(*this);
         return *this;
     }
 
@@ -344,7 +344,9 @@ public:
         base::reset();
     }
 
-    template <typename = std::enable_if<Copyable && !Movable>>
+    template <bool CopyableAlias = Copyable,
+              bool MovableAlias  = Movable,
+              typename           = std::enable_if_t<CopyableAlias && !MovableAlias>>
     void swap(any& rhs)
     {
         if (this == &rhs)
@@ -373,8 +375,8 @@ public:
         }
     }
 
-    template <typename = std::enable_if<Movable>>
-    void swap(any&& rhs) noexcept
+    template <bool MovableAlias = Movable, typename = std::enable_if_t<MovableAlias>>
+    void swap(any& rhs) noexcept
     {
         if (this == &rhs)
         {
--- a/include/cetl/any.hpp
+++ b/include/cetl/any.hpp
@@ -113,6 +113,7 @@ private:
 
     // Holds type-erased value destroyer. `nullptr` if storage has no value stored.
     void (*value_destroyer_)(void* self) = nullptr;
+
 };  // base_storage
 
 // Copy policy.
@@ -365,7 +366,7 @@ public:
             if (rhs.has_value())
             {
                 any tmp{rhs};
-                static_cast<base&>(rhs) = *this;
+                static_cast<base&>(rhs)   = *this;
                 static_cast<base&>(*this) = tmp;
             }
             else
@@ -394,7 +395,7 @@ public:
             if (rhs.has_value())
             {
                 any tmp{std::move(rhs)};
-                static_cast<base&>(rhs) = std::move(*this);
+                static_cast<base&>(rhs)   = std::move(*this);
                 static_cast<base&>(*this) = std::move(tmp);
             }
             else
--- a/include/cetl/any.hpp
+++ b/include/cetl/any.hpp
@@ -96,30 +96,6 @@ public:
         return get_raw_storage();
     }
 
-    void copy_from(const base_storage& src)
-    {
-        if (src.has_value())
-        {
-            copy_handlers_from(src);
-            assert(nullptr != value_copier_);
-
-            value_copier_(src.get_raw_storage(), get_raw_storage());
-        }
-    }
-
-    void move_from(base_storage& src) noexcept
-    {
-        if (src.has_value())
-        {
-            copy_handlers_from(src);
-            assert(nullptr != value_mover_);
-
-            value_mover_(src.get_raw_storage(), get_raw_storage());
-
-            src.reset();
-        }
-    }
-
     void copy_handlers_from(const base_storage& src) noexcept
     {
         value_destroyer_ = src.value_destroyer_;
@@ -162,11 +138,12 @@ struct base_copy<Footprint, true, Alignment> : base_storage<Footprint, Alignment
     constexpr base_copy() = default;
     base_copy(const base_copy& other)
     {
-        base::copy_from(other);
+        copy_from(other);
     }
 
     base_copy& operator=(const base_copy& other)
     {
+        base::reset();
         copy_from(other);
         return *this;
     }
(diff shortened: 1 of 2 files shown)


include/cetl/any.hpp
```diff
@@ -26,69 +26,226 @@ class any;
...
+        if (!has_value())
+        {
+            return nullptr;
+        }
+
+        // TODO: Add RTTI check here.
+        return get_raw_storage();
+    }
+
+    void copy_from(const base_storage& src)
```
pavel-kirienko, line 84: If `this` already contains a value, what happens to it? The type-erased copy handler does not (cannot really) invoke the dtor.

Same question for the move case.

include/cetl/any.hpp
```diff
@@ -26,69 +26,226 @@ class any;
...
+            copy_handlers_from(src);
+            assert(nullptr != value_mover_);
+
+            value_mover_(src.get_raw_storage(), get_raw_storage());
+
+            src.reset();
+        }
+    }
+
+    void copy_handlers_from(const base_storage& src) noexcept
```
pavel-kirienko, line 108: May I suggest that we avoid introducing unused data members by moving the type-erased copy and move handlers into the specializations of copy and move for which the corresponding operations are enabled? This would be similar to your telescoping `make_handlers` method where a derived class calls the base class.

This is not an optimization (nobody cares about the extra 8-16 bytes for the function pointers) but an attempt to use stronger typing and reduce the state space (no way to call a null function pointer if it's not there in the first place).

This change would affect:

- `copy_from` (only defined in `base_copy`; can be made private probably, see my comment about `swap`)
- `move_from` (only defined in `base_move`)
- `copy_handlers_from` (telescoping like `make_handlers`)
- `reset` (ditto)

include/cetl/any.hpp
```diff
@@ -105,46 +262,35 @@ template <std::size_t Footprint,
...
 
     template <
         typename ValueType,
         typename Tp = std::decay_t<ValueType>,
         typename = std::enable_if_t<!std::is_same<Tp, any>::value && !pf17::detail::is_in_place_type<ValueType>::value>>
     any(ValueType&& value)
     {
-        soo_handler<Tp>::create(*this, std::forward<ValueType>(value));
+        base::template make_handlers<Tp>();
+        new (base::get_raw_storage()) Tp(std::forward<ValueType>(value));
```
pavel-kirienko, line 277: The pattern with `make_handlers` and then placement `new` repeats a lot here; should we extract it into a method?

include/cetl/any.hpp
```diff
@@ -177,22 +323,31 @@ class any : private detail::base_move<Footprint, Copyable, Movable, Alignment>
...
+
+        base::template make_handlers<Tp>();
+        const auto ptr = new (base::get_raw_storage()) Tp(list, std::forward<Args>(args)...);
+        return *ptr;
     }
 
     void reset() noexcept
     {
-        handle(detail::action::Destroy);
+        base::reset();
```
pavel-kirienko, line 342: `using base::reset;`?

include/cetl/any.hpp
```diff
@@ -203,117 +358,63 @@ class any : private detail::base_move<Footprint, Copyable, Movable, Alignment>
         {
             if (rhs.has_value())
             {
-                any tmp;
-                rhs.handle(detail::action::Move, &tmp);
-                handle(detail::action::Move, &rhs);
-                tmp.handle(detail::action::Move, this);
+                any tmp{rhs};
+                rhs.copy_from(*this);
+                base::copy_from(tmp);
```
pavel-kirienko, line 361: Now your `copy_from` and `move_from` are no different from the standard copy/move assignment operators, why not invoke them directly here via `static_cast<base_copy&>`? This will allow you to make `copy_from` and `move_from` private (or remove them completely idk).

pavel-kirienko (changes requested):
- Are we going to introduce the `any_like` later? I just realized that in C++14, this make_any business would only work if `Copyable` is true, because in C++14 NRVO is not guaranteed.

- Should we add `CETL_NODISCARD` to all any_casts?

include/cetl/any.hpp
serges147: There are no such paths - these `copy_from` & `move_from` are always called ONLY from empty `this`. I'm gonna add such `assert`-s (like `assert(!has_value());`) to this functions.

serges147:
>Are we going to introduce the any_like later
Yes, right after RTTI business. Probably this `cetl::any` should be renamed to `any_like`, so that polyfill will have instead `any` as much as possible close to std behavior.

pavel-kirienko:
`any_like` was meant to be a type helper. I don't think renaming `cetl::any` into any_like is a good idea as the name doesn't really reflect the purpose.

include/cetl/any.hpp
serges147: >why not invoke them directly
yes, now I see (with help of `static_cast`) it's possible. previously I though that recursion will prevent it - haven't thought about casting

include/cetl/any.hpp
serges147: Also, after addressing below comment...
namely: _> This will allow you to make copy_from and move_from private..._
both of these function are now private and called only 2 times (previously 5), so `this` is definitely empty at this point.

serges147: Actually, I was wrong about `base_copy& operator=(const base_copy& other)` ! There is indeed memory leak in case `this` already contains a value. I'm extending tests with simple verification that number constructs equals to number of destructs in the end.

pavel-kirienko (commented):
I've looked at the code again and there seems to be no new feedback from me at the moment.

I pushed a fix to correct the test that used to fail in release builds.

include/cetl/any.hpp
serges147: Please see new intermediate `base_handlers`.

serges147:
>I pushed a fix to correct the test
omg, it was silly of me ;) - although it would be discovered if I had rtti ready - working on that (for next pr)...

pavel-kirienko (changes requested):
Ok nice, now we need to specialize base_handlers for Copyable and Movable, four combinations overall. It is also better to keep member variables toward the end of class definitions for consistency.

serges147:
> Ok nice, now we need to specialize base_handlers for Copyable and Movable, four combinations overall. It is also better to keep member variables toward the end of class definitions for consistency.

Done.

include/cetl/any.hpp
```diff
@@ -26,69 +26,302 @@ class any;
...
+    void reset() noexcept
+    {
+        value_copier_ = nullptr;
+        base::reset();
+    }
+
+    // Holds type-erased value copyer. `nullptr` when copy operation is not supported.
+    void (*value_copier_)(const void* src, void* dst) = nullptr;
+
+private:
```
pavel-kirienko, line 135: Here and below, I don't really see the point of this specifier; I think it should be removed.

pavel-kirienko (approved):

PR #110, opened 2024-03-27
Use CETL's polymorphic RTTI

serges147:
- Now CETL's polymorphic RTTI casting is in use at `cetl::detail::base_storage`.
- Extended unit tests `TestXxx` helper types with virtual `what()` method - to test polymorphic RTTI casting.
- Added compile tests to verify several `static_assert`-s at `cetl::any` (related to `Footprint` fitting).

Also:
- Reduced code duplication at unit tests `TestXxx` helper types.
- Fixed incorrect applying of `-fno-exceptions` to C flags - should be C++ flags.
- Addressed several unit tests todo-s. No todo-s anymore at production `cetl::any` code.
- Added verification that there will be no value moved out of source `any` in case of bad cast attempt.

This change was also made, but decided to revert (for now) and save it as patch (for future reference):
- Added new `CETLVAST_DISABLE_CPP_RTTI` configuration parameter. `OFF` by default.
[CETLVAST_DISABLE_CPP_RTTI.patch](https://github.com/OpenCyphal/CETL/files/14792699/CETLVAST_DISABLE_CPP_RTTI.patch)

### DIFF
--- a/include/cetl/any.hpp
+++ b/include/cetl/any.hpp
@@ -8,6 +8,7 @@
 #ifndef CETL_ANY_HPP_INCLUDED
 #define CETL_ANY_HPP_INCLUDED
 
+#include "rtti.hpp"
 #include "pf17/cetlpf.hpp"
 #include "pf17/utility.hpp"
 #include "pf17/attribute.hpp"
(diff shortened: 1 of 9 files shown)


include/cetl/any.hpp
```diff
@@ -49,41 +53,97 @@ struct base_storage  // NOLINT(*-pro-type-member-init)
...
         value_destroyer_ = [](void* const storage) {
             const auto ptr = static_cast<Tp*>(storage);
             ptr->~Tp();
         };
+
+        make_converters<Tp>();
+
+        value_type_informer_ = [](const void*) {
+
+#if __cpp_rtti
```
pavel-kirienko, line 72: Here, we are stepping on a slippery slope of relying on implementation-specific features. In general, as far as I know, the C++ standard provides no feature test macro for RTTI:

image

I would limit our implementation of `any` to CETL RTTI only.

From our earlier conversations, I understood that you were adding the type informer handler `value_type_informer_` in order to support CETL RTTI rather than the standard `typeid`.

include/cetl/any.hpp
```diff
@@ -49,41 +53,97 @@ struct base_storage  // NOLINT(*-pro-type-member-init)
...
+        CETL_DEBUG_ASSERT(nullptr != value_type_informer_, "");
+
+        return static_cast<const std::type_info*>(value_type_informer_(get_raw_storage()));
     }
+#endif
 
     template <typename ValueType>
     CETL_NODISCARD void* get_ptr() noexcept
     {
+        static_assert(sizeof(ValueType) <= Footprint, "Enlarge the footprint");
```
pavel-kirienko, line 116: I think it would be more accurate to say that the type cannot contain the requested type by its definition since the footprint would be too small. The suggestion to enlarge the footprint is only meaningful in a context where one is attempting to place a large object into the `any` container.

Same comment applies below.

include/cetl/any.hpp
serges147: ok, I didn't know that `__cpp_rtti` is some kind of an extension. I will remove related stuff and stick with our CETL RTTI only. Could you please give a link (or pdf) from where you made the screenshot - for future reference. Thnx!

include/cetl/any.hpp
serges147: ok, you are right, I'll change them to "Cannot contain the requested type since the footprint is too small".

include/cetl/any.hpp
pavel-kirienko: here: https://github.com/cplusplus/draft/releases/tag/n4917

pavel-kirienko (approved):

pavel-kirienko:
I think it's fine overall but there's one case that should be working but it's not; there's a missing remove_cv somewhere:

```c++
#include <cetl/any.hpp>
#include <iostream>
#include <string>


namespace cetl
{
template <>
constexpr type_id type_id_value<int> = {1,2,3,4,5};
template <>
constexpr type_id type_id_value<std::string> = {1,2,3,4,6};
}


int main()
{
    cetl::any<64> any;
    any.emplace<std::string>("hello");
    std::cout << *cetl::any_cast<std::string>(&any) << std::endl;
    std::cout << *cetl::any_cast<const std::string>(&any) << std::endl;  // here
}
```

serges147:
_>should be working but it's not_

I'll take a look at this

Issue #118, opened 2024-04-22
Enhance `cetl::pmr::Factory::make_unique` to support "interface/concrete" scenario.


serges147:
1. Currently you have to do something like this:
```
template <typename Tag, typename... Args>
CETL_NODISCARD UniquePtr<typename Tag::Interface> makeUniquePtr(cetl::pmr::memory_resource& memory, Args&&... args)
{
    PmrAllocator<typename Tag::Concrete> allocator{&memory};
    auto interface_deleter = typename UniquePtr<typename Tag::Interface>::deleter_type{allocator, 1};

    auto concrete  = cetl::pmr::Factory::make_unique(allocator, std::forward<Args>(args)...);
    auto interface = UniquePtr<typename Tag::Interface>{concrete.release(), interface_deleter};

    return interface;
}
```
2. The above code has issue with `cetl::pmr::memory_resource`, namely incorrect `size` on deallocation (8 bytes instead of real/concrete size which was during `allocate`).

serges147:
I believe this is done in context of PR # 124 - the new `cetl::pmr::InterfaceFactory` and its `cetl::pmr::InterfacePtr`.
@pavel-kirienko would you agree?

PR #123, opened 2024-05-15
Issue #118: Enhance `cetl::pmr::Factory::make_unique` to support "interface/concrete" scenario

serges147:


cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
```diff
@@ -148,3 +231,34 @@ TEST(example_07_polymorphic_alloc_deleter, example_usage_1)
...
     //![example_usage_1]
 }
+
+TEST_F(example_07_polymorphic_alloc_deleter, example_usage_2)
+{
+    //![example_usage_2]
+
+    cetl::pmr::polymorphic_allocator<MyObject> alloc{cetl::pmr::new_delete_resource()};
+
+    auto obj0 = cetl::pmr::InterfaceFactory::make_unique<MyObject>(alloc, "obj1", 4U);
```
serges147, line 241: This is as it was before - we make `MyObject` instance, and return `MyObject` "interface".

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
```diff
@@ -148,3 +231,34 @@ TEST(example_07_polymorphic_alloc_deleter, example_usage_1)
...
+
+    auto obj1 = cetl::pmr::InterfaceFactory::make_unique<IIdentifiable>(alloc, "obj1", 4U);
```
serges147, line 243: This is new - we make `MyObject` instance as before, BUT we return `IIdentifiable` interface. Note that 'IIdentifiable` has `protected` destructor.

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
```diff
@@ -148,3 +231,34 @@ TEST(example_07_polymorphic_alloc_deleter, example_usage_1)
...
+    {
+        std::cout << "Obj1 id  : " << obj1->id() << std::endl;
+        obj1.reset();
+        std::cout << std::endl;
+    }
+    auto obj2 = cetl::pmr::InterfaceFactory::make_unique<IDescribable>(alloc, "obj2", 4U);
```
serges147, line 249: Here I demonstrate conversion (`move` actually!) of `InterfacePtr<IDescribable>` → to compatible `InterfacePtr<INameable>`.

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,92 @@
...
+#define CETL_PMR_INTERFACE_PTR_H_INCLUDED
+
+#ifndef CETL_H_ERASE
+#    include "cetl/cetl.hpp"
+#endif
+
+#include "cetl/unbounded_variant.hpp"
+
+#include <memory>
+#include <functional>
```
serges147, line 16: This is temporary (see below `std::function<void(Interface*)> deleter_;`) - will be replaced with PMR `cetl::pmr::function` when it will be available.

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+namespace cetl
+{
+namespace pmr
+{
+
+template <typename Interface>
+class PmrInterfaceDeleter final
```
serges147, line 24: Note - there is no mention of `PmrAllocator` or its nested `value_type` (aka Concrete) types in the template params. This is crucial to make deleter "concrete type erased".

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,92 @@
...
+            using Concrete     = typename PmrAllocator::value_type;
+            auto* concrete_ptr = static_cast<Concrete*>(ptr);
+
+            concrete_ptr->~Concrete();
+            alloc.deallocate(concrete_ptr, obj_count);
+        };
+    }
+
+    template <typename Down>
+    PmrInterfaceDeleter(const PmrInterfaceDeleter<Down>& other)
```
serges147, line 40: This is important to make possible up-casting of `InterfacePtr<B>` → `InterfacePtr<A` (assuming `class B : A {...}` inheritance).

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,92 @@
...
+        {
+            deleter_(ptr);
+        }
+    }
+
+private:
+    template <typename Down>
+    friend class PmrInterfaceDeleter;
+
+    std::function<void(Interface*)> deleter_;
```
serges147, line 60: Will be replaced with PMR `cetl::pmr::function` when it will be available.

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
```diff
@@ -9,43 +9,98 @@
...
+#include "cetl/pmr/interface_ptr.hpp"
 #include "cetl/pmr/memory.hpp"
 
 #include <unordered_map>
 #include <iostream>
-#include <string.h>
 
 #include <gtest/gtest.h>
 
+class INameable
```
serges147, line 20: In the end of this file I introduce new "example_usage_2" test case - this is where I demonstrate the "interface"-part feature of new `cetl::pmr::InterfacePtr`. So, I need couple interfaces for the `MyObject` final class.

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
```diff
@@ -148,3 +228,34 @@ TEST(example_07_polymorphic_alloc_deleter, example_usage_1)
...
     //![example_usage_1]
 }
+
+TEST_F(example_07_polymorphic_alloc_deleter, example_usage_2)
+{
+    //![example_usage_2]
+
+    cetl::pmr::polymorphic_allocator<MyObject> alloc{cetl::pmr::new_delete_resource()};
+
+    auto obj0 = cetl::pmr::InterfaceFactory::make_unique<MyObject>(alloc, "obj0", 4U);
```
serges147, line 238: This is as it was before (with `cetl::pmr::Factory`) - we make `MyObject` instance, and return `MyObject` "interface".

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
pavel-kirienko: `INameable` implies that you can give it a name or use `name` as a verb, no? If it has a name that can't be changed then shouldn't it be `INamed`? Then the second one would be `IDescribed` (or you'd need to change `description()` into `describe()`). I think there is no problem with `IIdentifiable` because one could argue that it uses `id` as a verb.

Normally I would consider it a nitpick but this is a doc example where names matter.

include/cetl/pmr/interface_ptr.hpp
pavel-kirienko: Note for Scott & Lydia: `cetl::function` is a near-term goal that we are now working on.

include/cetl/pmr/interface_ptr.hpp
pavel-kirienko: 1. Upcasting as you described is unsafe to do via `static_cast` in the presence of data members. Why do we need to allow this conversion?

2. This definition hides the default copy and move constructors. I think it needs either sfinae or simply explicitly defaulted copy/move ctors (as they should take precedence over the template).

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,90 @@
...
+{
+namespace pmr
+{
+
+template <typename Interface>
+class PmrInterfaceDeleter final
+{
+public:
+    template <typename PmrAllocator>
+    PmrInterfaceDeleter(PmrAllocator alloc, std::size_t obj_count)
```
pavel-kirienko, line 26: Why are you passing the allocator by value here and also capturing it by value?

thirtytwobits (commented):
Looking good!

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
serges147: I believe there is no such notion that "xxx-able" is necessarily has some kind of "setter" logic. IMO all it says that there is a "xxx" verb which can be applied to an entity. In your example with "id" it looks like it is just shortcut for "identify" verb.

Also, in my experience past tense suffix (-ed) is commonly in use to connote "returning a new instance of something transformed", like const `arr.sorted()` vs in-place `arr.sort()` mutation.

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
serges147: No, here we have the opposite, namely upcasting (internally) from derived (aka Concrete) to a base class (aka Interface).

include/cetl/pmr/interface_ptr.hpp
serges147: > 1. Upcasting as you described is unsafe to...

Please correct me if I'm wrong, but I think this is actually DOWN-casting, and yes it's not safe generally. I need it to return back to the original pointer to concrete type (before calling its destructor and following deallocate at PMR). Without it following won't work:
```
class Concrete : Base /*sizeof f.e. 32*/, pubic ISomething
{
}
```
Instance of `InterfacePtr<ISomething>` (originally moved from `InterfacePtr<Concrete>`) stores internally pointer which is +32 bytes offset comparing to the original memory allocated (and initialized) for `Concrete` type instance.
So, when we are ready delete it then we need return back to the original pointer.

include/cetl/pmr/interface_ptr.hpp
serges147: yes, I already have my private `sshirokov/84_pmr_function` branch with very first draft of implementation.

include/cetl/pmr/interface_ptr.hpp
serges147: As far as I see (at std, and llvm & gcc implementation) it's common practice to pass allocators and deleters by value. I think this is done to eliminate "ownership" and lifetime issues.

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
```diff
@@ -148,3 +228,34 @@ TEST(example_07_polymorphic_alloc_deleter, example_usage_1)
 
     //![example_usage_1]
 }
+
+TEST_F(example_07_polymorphic_alloc_deleter, example_usage_2)
+{
+    //![example_usage_2]
+
+    cetl::pmr::polymorphic_allocator<MyObject> alloc{cetl::pmr::new_delete_resource()};
```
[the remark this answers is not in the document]
serges147: @thirtytwobits I see in other examples a lot of `new_delete_resource` calls. Are you suggesting to refactor a bit them as well?

Also, could you please take a look at `cetl/pf17/cetlpf.hpp` - it looks like we forgot to add inline `get_default_resource`-s there (similar to what we have for `new_delete_resource`).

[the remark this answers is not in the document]
thirtytwobits: It was only a nit. I don't really care. 

Yeah, I think you're right about adding this to the polyfill header.

cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
pavel-kirienko: Oh, I understand it now: internally, it is implemented as `static_cast<INamed>(my_object)` rather than `static_cast<INamed>(describable)`, that makes sense.

include/cetl/pmr/interface_ptr.hpp
pavel-kirienko: Okay; I thought about pmr semantics here but it doesn't really apply. It is not appropriate to pass a pmr by value since it's polymorphic, but this is just an allocator. This raises another question: does it really need to be a PMR allocator? The code is generic and should be equally applicable to an allocator that does not use PMR, so do we need `Pmr` in the name?

serges147: "Pmr" here b/c in reality it will be `cetl::pmr::polymorphic_allocator<Concrete>`, and its `deallocate` method will be in use on interface ptr ~destructor (or `reset`). Also, allocator itself is not polymorphic - internally used `cetl::pmr::memory_resourse` is.

serges147:
Closing this PR to create a new, production one.

PR #124, opened 2024-05-28
PMR support for `cetl::unbounded_variant` & `cetl::pmr::function`.

serges147:
- Added `IsPmr` support to `cetl::unbounded_variant`.
- Added new `cetl::pmr::function` (aka `std::function` but with PMR instead of c++ heap).
- Added new `cetl::pmr::InterfacePtr` and its `InterfaceFactrory::make_unique`
  - based on `std::unique_ptr` but with custom PMR-based allocator and deleter
  - allows to hide implementation (aka Concrete) type from result InterfacePtr
  - allows to delete destructor from an interface type

### DIFF
--- a/cetlvast/suites/unittest/test_unbounded_variant.cpp
+++ b/cetlvast/suites/unittest/test_unbounded_variant.cpp
@@ -273,8 +273,6 @@ struct Empty
 class TestPmrUnboundedVariant : public testing::Test
 {
 protected:
-    void SetUp() override {}
-
     void TearDown() override
     {
         EXPECT_THAT(mr_.allocations, IsEmpty());
@@ -1172,12 +1170,14 @@ TEST_F(TestPmrUnboundedVariant, pmr_ctor)
     EXPECT_THAT(dst.has_value(), false);
     EXPECT_THAT(dst.get_memory_resource(), get_mr());
 
-    dst = ub_var{'x'};
+    dst = ub_var{get_mr(), 'x'};
     EXPECT_THAT(dst.has_value(), true);
     EXPECT_THAT(get<char>(dst), 'x');
+    EXPECT_THAT(dst.get_memory_resource(), get_mr());
 
     dst = Empty{};
     EXPECT_THAT(dst.has_value(), true);
+    EXPECT_THAT(dst.get_memory_resource(), get_mr());
 
     ub_var dst2{};
     EXPECT_THAT(dst2.get_memory_resource(), cetl::pmr::get_default_resource());
@@ -1187,6 +1187,7 @@ TEST_F(TestPmrUnboundedVariant, pmr_ctor)
 
     dst2 = {};
     EXPECT_THAT(dst2.has_value(), false);
+    dst2.set_memory_resource(get_mr());
 
     dst2 = std::uint16_t{0x147};
     EXPECT_THAT(dst2.has_value(), true);
@@ -1210,9 +1211,13 @@ TEST_F(TestPmrUnboundedVariant, pmr_ctor)
     EXPECT_THAT(dst4.has_value(), false);
     EXPECT_THAT(dst4.get_memory_resource(), get_mr());
 
-    const ub_var dst5{std::move(dst4)};
+    ub_var dst5{std::move(dst4)};
     EXPECT_THAT(dst5.has_value(), false);
     EXPECT_THAT(dst5.get_memory_resource(), get_mr());
+
+    dst5 = {};
+    EXPECT_THAT(dst5.has_value(), false);
+    EXPECT_THAT(dst5.get_memory_resource(), cetl::pmr::get_default_resource());
 }
 
 }  // namespace
--- a/cetlvast/suites/unittest/test_unbounded_variant.cpp
+++ b/cetlvast/suites/unittest/test_unbounded_variant.cpp
@@ -1093,7 +1093,8 @@ TEST_F(TestPmrUnboundedVariant, emplace_1)
         using ub_var = unbounded_variant<sizeof(char)>;
 
         ub_var src;
-        src.emplace<char>('Y');
+        auto   y_ptr = src.emplace<char>('Y');
+        EXPECT_THAT(get_if<char>(&src), y_ptr);
         EXPECT_THAT(get<char>(src), 'Y');
     }
 
@@ -1113,10 +1114,43 @@ TEST_F(TestPmrUnboundedVariant, emplace_1)
         using ub_var = unbounded_variant<sizeof(MyType)>;
 
         ub_var t;
-        t.emplace<MyType>('Y', 147);
+        auto   my_ptr = t.emplace<MyType>('Y', 147);
+        EXPECT_THAT(get_if<MyType>(&t), my_ptr);
         EXPECT_THAT(get<MyType>(t).ch_, 'Y');
         EXPECT_THAT(get<MyType>(t).number_, 147);
     }
+}
+
+TEST_F(TestPmrUnboundedVariant, emplace_1_ctor_exception)
+{
+#if defined(__cpp_exceptions)
+
+    side_effect_stats stats;
+    {
+        using ub_var = unbounded_variant<sizeof(MyCopyableAndMovable)>;
+
+        auto stats_side_effects    = stats.make_side_effect_fn();
+        auto throwing_side_effects = [=](side_effect_op op) {
+            stats_side_effects(op);
+            if (op == side_effect_op::Construct)
+            {
+                throw std::runtime_error("ctor");
+            }
+        };
+
+        ub_var t;
+        EXPECT_THAT(t.has_value(), false);
+        EXPECT_THAT(t.valueless_by_exception(), false);
+
+        EXPECT_THROW(sink(t.emplace<MyCopyableAndMovable>('Y', throwing_side_effects)), std::runtime_error);
+    }
+    EXPECT_THAT(stats.constructs, 1);
+    EXPECT_THAT(stats.destructs, 0);
+    EXPECT_THAT(stats.ops, "@");
+
+#else
+    GTEST_SKIP() << "Not applicable when exceptions are disabled.";
+#endif
 }
 
 TEST_F(TestPmrUnboundedVariant, emplace_2_initializer_list)
--- a/cetlvast/suites/unittest/test_unbounded_variant.cpp
+++ b/cetlvast/suites/unittest/test_unbounded_variant.cpp
@@ -36,7 +36,6 @@ using testing::Return;
 using testing::IsNull;
 using testing::IsEmpty;
 using testing::NotNull;
-using testing::InSequence;
 using testing::StrictMock;
 
 using namespace std::string_literals;
--- a/cetlvast/suites/unittest/test_unbounded_variant.cpp
+++ b/cetlvast/suites/unittest/test_unbounded_variant.cpp
@@ -265,6 +265,8 @@ private:
     using base = TestBase;
 };
 
+struct Empty {};
+
 /// TESTS -----------------------------------------------------------------------------------------------------------
 
 TEST(test_unbounded_variant, bad_unbounded_variant_access_ctor)
@@ -1099,14 +1101,19 @@ TEST(test_unbounded_variant, emplace_2_initializer_list)
 
 TEST(test_unbounded_variant, pmr_only_ctor)
 {
-    using ub_var = unbounded_variant<0, true, true, 1, true>;
+    using ub_var =
+        unbounded_variant<0 /*Footprint*/, true /*Copyable*/, true /*Movable*/, 1 /*Alignment*/, true /*IsPmr*/>;
 
-    auto mr = cetl::pmr::get_default_resource();
-
-    ub_var dst{mr};
+    ub_var dst{};
     EXPECT_THAT(dst.has_value(), false);
 
-    dst = ub_var{mr, 'x'};
+    dst = ub_var{'x'};
+
+    dst = Empty{};
+
+    ub_var dst2{};
+    dst2 = std::move(dst);
+    dst2 = {};
 }
 
 }  // namespace
@@ -1143,6 +1150,9 @@ constexpr type_id type_id_value<std::complex<double>> = {8};
 
 template <>
 constexpr type_id type_id_value<std::function<const char*()>> = {9};
+
+template <>
+constexpr type_id type_id_value<Empty> = {10};
 
 }  // namespace cetl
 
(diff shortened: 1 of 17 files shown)


cetlvast/suites/unittest/test_unbounded_variant.cpp
```diff
@@ -598,36 +663,36 @@ TEST(test_unbounded_variant, assign_1_copy)
...
 
         EXPECT_THAT(get<const test&>(src2).value_, 10);
         EXPECT_THAT(get<const test&>(src2).payload_, 'Z');
-        EXPECT_THAT(get<const test&>(dst).value_, 30);
+        EXPECT_THAT(get<const test&>(dst).value_, 20);
         EXPECT_THAT(get<const test&>(dst).payload_, 'Z');
     }
     EXPECT_THAT(stats.constructs, stats.destructs);
-    EXPECT_THAT(stats.ops, "@CCC~@CCC~C~C~~~~~~~");
+    EXPECT_THAT(stats.ops, "@CC@C~C~~~~~");
```
serges147, line 692: All these shortened `stats.ops` (f.e.
`"@CCC~@CCC~C~C~~~~~~~"` →
`"@CC@C~C~~~~~"`
) are result of eliminating of the "swap" pattern from `unbounded_variant::operator=`-s.

cetlvast/suites/unittest/test_unbounded_variant.cpp
```diff
@@ -79,53 +96,53 @@ struct side_effect_stats
     }
 };
 
-struct TestBase : rtti_helper<type_id_type<0x0>>
+struct MyBase : rtti_helper<type_id_type<0x0>>
```
serges147, line 99: Renamed these helper classes to not be confused with actual `TestXxx` gtest fixture class (like `TestPmrUnboundedVariant` is).

include/cetl/pmr/function.hpp
```diff
@@ -0,0 +1,321 @@
...
+            any_handler_       = std::move(other.any_handler_);
+            handler_ptr_       = get_if<detail::function_handler<Result, Args...>>(&any_handler_);
+            other.handler_ptr_ = nullptr;
+
+            CETL_DEBUG_ASSERT(any_handler_.has_value(), "");
+            CETL_DEBUG_ASSERT(nullptr != handler_ptr_, "");
+        }
+    }
+
+    // TODO: Add Callable constraint;
```
serges147, line 118: These "Add Callable constraint" todos will be addressed separately. I believe it can be properly done by adding `cetl::pf17::invoke` support (like `cetl::is_invokable<...>` stuff).

include/cetl/unbounded_variant.hpp
```diff
@@ -48,36 +48,461 @@ class bad_unbounded_variant_access : public std::bad_cast
...
 
 namespace detail
 {
 
+inline std::nullptr_t throw_bad_alloc()
+{
+#if defined(__cpp_exceptions)
+    throw std::bad_alloc();
+#else
+    return nullptr;
```
serges147, line 62: We don't terminate here but instead return `nullptr` which makes unbounded variant as "valueless by exception".

include/cetl/unbounded_variant.hpp
```diff
@@ -48,36 +48,461 @@ class bad_unbounded_variant_access : public std::bad_cast
...
+
+[[noreturn]] inline void throw_bad_unbounded_variant_access()
+{
+#if defined(__cpp_exceptions)
+    throw bad_unbounded_variant_access();
+#else
+    CETL_DEBUG_ASSERT(false, "bad_unbounded_variant_access");
+    std::terminate();
+#endif
+}
```
serges147, line 74: This method has not been changed - just moved a bit up.

include/cetl/unbounded_variant.hpp
```diff
@@ -520,10 +1091,9 @@ class unbounded_variant : detail::base_move<Footprint, Copyable, Movable, Alignm
     /// \param args Arguments to be forwarded to the constructor of `ValueType`.
     ///
     template <typename ValueType, typename... Args, typename Tp = std::decay_t<ValueType>>
-    Tp& emplace(Args&&... args)
+    Tp* emplace(Args&&... args)
```
serges147, line 1094: Now we support possible PMR failures, so in case c++ are off you get `nullptr` result from `emplace`).

include/cetl/pmr/function.hpp
```diff
@@ -0,0 +1,321 @@
...
+
+#include <cstddef>
+#include <exception>
+#include <functional>
+
+namespace cetl
+{
+namespace pmr
+{
+
```
serges147, line 24: This PR is more concentrated about `unbounded_variant` and adding PMR (and `valueless_by_exception`!).
That is why `function` and `InterfacePtr` not **YET** properly documented, and don't have yet proper error handling/indication - these will be addressed in the next PR (please).

cetlvast/suites/unittest/test_unbounded_variant.cpp
serges147: BTW, elimination of the "swap" pattern is needed to properly maintain the new "valueless by exception" state. Otherwise the state will end up at some temporary swap var only, and won't be properly passed/reflected on actual target objects.

include/cetl/unbounded_variant.hpp
```diff
@@ -48,36 +48,463 @@ class bad_unbounded_variant_access : public std::bad_cast
...
+        CETL_DEBUG_ASSERT(0UL == value_size_, "This object is expected to be a brand new one.");
+
+        // We need to store (presumably) allocated size b/c this is important
+        // for detection of the "valueless by exception" state in case something goes wrong later
+        // (like in-place value construction failure) - in such failure state
+        // there will `>0` value size BUT `nullptr` handlers (f.e. `value_destroyer_`).
+        //
+        value_size_ = size_bytes;
+
+        return inplace_buffer_;
```
thirtytwobits, line 103: should we not check that size_bytes is <= Footprint and return null if not?

thirtytwobits (commented):
We also have a cetlpf.hpp in unbounded_variant.hpp we need to remove.

Will use of "cetl/unbounded_variant.hpp" compile if IsPmr is false and no pmr types are included? I would expect this. That is, if we don't require PMR then we should require inclusion of PMR types.

serges147:
> We also have a cetlpf.hpp in unbounded_variant.hpp we need to remove.

Sorry Scott, but when it comes to "include or not to include" of these two:
- `cetl/pfNN/cetlpf.hpp`
- `#ifndef CETL_H_ERASE #include "cetl/cetl.hpp" #endif`
I'm quite confused when which... so far I tried to do as it's in other files.
Also, at `cetl/unbounded_variant.hpp` I use `cetl::byte` (which is either `std::byte` or our `cetl::pf17::byte` depending on c++ standard), so how can I use `cetl::byte` without including first `cetl/pf17/cetlpf.hpp` ?


> Will use of "cetl/unbounded_variant.hpp" compile if IsPmr is false and no pmr types are included? I would expect this. That is, if we don't require PMR then we should require inclusion of PMR types.

pmr types are included regardless of IsPmr. I believe it's inevitable (`IsPmr` is template param, not like some `#define`). Having said this, You don't pay for what you don't use - all these PMR-specific methods should not be instantiated in the first place (and there are `static_assert`-s to ensure you don't call them by mistake).

include/cetl/unbounded_variant.hpp
serges147: No, for non-pmr there is no such thing. Please see static assert at `make_handlers` - this method is one of the core entry points when it come to creation of a new unbounded variable. So it won't compile (and there is compile test to ensure this, see `test_unbounded_variant_footprint_set.cpp`).

There are also two other static asserts on attempt to read an arbitrary value (see `get_ptr` overloads, and corresponding `test_unbounded_variant_footprint_get_[non_]const.cpp` compile tests).

thirtytwobits:
> 
> Sorry Scott, but when it comes to "include or not to include" of these two...
> 
> * `cetl/pfNN/cetlpf.hpp`
> * `#ifndef CETL_H_ERASE #include "cetl/cetl.hpp" #endif`
> 
> ... I'm quite confused when which... so far I tried to do as it's in other files. Also, at `cetl/unbounded_variant.hpp` I use `cetl::byte` (which is either `std::byte` or our `cetl::pf17::byte` depending on c++ standard), so how can I use `cetl::byte` without including first `cetl/pf17/cetlpf.hpp` ?

the polyfill headers are completely optional in CETL so nothing in CETL should include them, they should include other things from CETL. As for byte, you'll just have to use uint8_t internally. I don't think we're providing `byte` as part of a public API anywhere in here.

The idea CETL type only includes `cetl.hpp` but there are some exceptions to this rule; just never polyfill headers is all.

thirtytwobits:
> pmr types are included regardless of IsPmr. I believe it's inevitable (`IsPmr` is template param, not like some `#define`). Having said this, You don't pay for what you don't use - all these PMR-specific methods should not be instantiated in the first place (and there are `static_assert`-s to ensure you don't call them by mistake).

If PMR types are always required then we shouldn't also need the "isPMR" parameter. Instead we should just model the default storage as a memory_resource we define locally.

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,92 @@
...
+                       concrete_ptr->~Concrete();
+                       alloc.deallocate(concrete_ptr, obj_count);
+                   }}
+    {
+    }
+
+    template <typename Down, typename = std::enable_if_t<std::is_base_of<Interface, Down>::value>>
+    PmrInterfaceDeleter(const PmrInterfaceDeleter<Down, Pmr>& other)
+        : deleter_{other.get_memory_resource(), [other](Interface* ptr) {
+                       // Delegate to the down class deleter.
```
pavel-kirienko, line 38:
```suggestion
                       // Delegate to the down class deleter.
                       // The down-casting is assumed to be safe because both the base type
                       // and the derived type are statically known in this context.
                       // This is a possible deviation from AUTOSAR M5-2-3; whether the
                       // type is polymorphic or not is irrelevant in this context.
```

pavel-kirienko (dismissed):
As we discussed, it is desirable to eliminate the use of PMR from the deleter, or at least (this may be a better option) provide a new non-PMR deleter that does not have the up-converting constructor and which is used by default while allowing the user to switch to the other one in case of great necessity.

include/cetl/pmr/interface_ptr.hpp
```diff
@@ -0,0 +1,194 @@
...
+    ///
+    void operator()(Interface* ptr) noexcept
+    {
+        deleter_(ptr);
+    }
+
+    // Below convertor constructor is only possible with enabled PMR at `function`.
+    // For now, we decided to comment it out, so that `InterfacePtr` always stays
+    // within 24-bytes small object optimization, namely without extra memory allocation
+    // just for the sake of "advanced" deleter (actually chain of casters down to original `Concrete` pointer).
```
pavel-kirienko, line 60:
```suggestion
    // just for the sake of "advanced" deleter (actually chain of casters down to original `Concrete` pointer).
    //
    // NOTE: It is possible to avoid PMR if we define a certain static limit on the maximum number of
    // down-conversions performed on an interface ptr. One issue is that exceeding that limit is a
    // runtime error, so it will have to be handled correctly somehow. It could be that std::abort()
    // is a sensible choice here, but probably not. The way to implement this solution is to keep a
    // stack of the following converters, where each converter reverses one up-conversion:
    //
    //  using caster_t = void* (*) (void*);
    //  /// Creates a function that performs a safe type conversion on a type-erased void* pointer.
    //  template <typename From, typename To>
    //  static caster_t make_caster() noexcept
    //  { return [](void* const ptr) -> void* { return static_cast<To*>(static_cast<From*>(ptr)); }; }
    //
```

include/cetl/pmr/function.hpp
```diff
@@ -0,0 +1,359 @@
...
+/// Exchanges the state of `lhs` with that of `rhs`. Effectively calls `lhs.swap(rhs)`.
+///
+/// Any failure during the swap operation could result in the "valueless by exception" state,
+/// and depending on which stage of swapping the failure happened
+/// it could affect (invalidate) either of `lhs` or `rhs` function.
+/// Use `valueless_by_exception()` method to check if a function is in such failure state,
+/// and `reset` (or `reset(Pmr*)`) method to recover from it.
+///
+template <typename Result, typename... Args, std::size_t Footprint, typename Pmr = void>
+inline void swap(cetl::pmr::function<Result(Args...), Footprint, Pmr>& lhs,
```
pavel-kirienko, line 351: `inline` is redundant for function templates, they are inline by default.

include/cetl/pmr/function.hpp
```diff
@@ -0,0 +1,359 @@
...
+{
+
+/// @brief Overloads the `std::swap` algorithm for `cetl::pmr::function`.
+/// Exchanges the state of `lhs` with that of `rhs`. Effectively calls `lhs.swap(rhs)`.
+///
+/// Any failure during the swap operation could result in the "valueless by exception" state,
+/// and depending on which stage of swapping the failure happened
+/// it could affect (invalidate) either of `lhs` or `rhs` function.
+/// Use `valueless_by_exception()` method to check if a function is in such failure state,
+/// and `reset` (or `reset(Pmr*)`) method to recover from it.
```
pavel-kirienko, line 348:
```suggestion
/// and `reset` (or `reset(Pmr*)`) method (or assign a new value) to recover from it.
```

include/cetl/pmr/function.hpp
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    void swap(function& other) noexcept
+    {
+        any_handler_.swap(other.any_handler_);
+
+        handler_ptr_       = get_if<handler_t>(&any_handler_);
+        other.handler_ptr_ = get_if<handler_t>(&other.any_handler_);
+    }
+
+    /// True if the function is valueless b/c of an exception.
```
pavel-kirienko, line 306:
```suggestion
    /// True if the function is valueless b/c of an exception or OOM.
```

OOM doesn't necessarily cause an exception

include/cetl/unbounded_variant.hpp
```diff
@@ -48,36 +44,502 @@ class bad_unbounded_variant_access : public std::bad_cast
...
+    {
     }
 
     CETL_NODISCARD bool has_value() const noexcept
     {
-        return nullptr != value_destroyer_;
+        return (base::get_value_size() > 0UL) && (nullptr != value_destroyer_);
+    }
+
+    /// True if the variant is valueless b/c of an exception.
```
pavel-kirienko, line 528:
```suggestion
    /// True if the variant is valueless b/c of an exception or OOM.
```

pavel-kirienko (dismissed):

pavel-kirienko (approved):

thirtytwobits (approved):

PR #126, opened 2024-06-03
get rid of inclusion of pf17 inside of `unbounded_variant`

serges147:


include/cetl/unbounded_variant.hpp
```diff
@@ -961,6 +961,57 @@ struct base_move<Pmr, Footprint, Copyable, true /*Movable*/, Alignment>
...
+
+/// Implementation similar to \ref std::in_place_type_t or \ref cetl::pf17::in_place_type_t.
+///
+/// In use by several \ref cetl::unbounded_variant constructors.
+///
+/// Can't use directly either of already existing `std::in_place_type_t` or `cetl::pf17::in_place_type_t` types due to
+/// C++14 limitation and polyfill optionality (by design in CETL, according to Scott), so a bit of code duplication.
+///
+template <typename T>
+struct in_place_type_t
```
thirtytwobits, line 979: Can this be a public member of unbounded_variant? `ub_var` feels like an ugly hack.

serges147: ok

pavel-kirienko (commented):
If we're going to keep this, should we not alert the user in the doc comments for these constructors that one should consider using the freestanding factory function instead:

https://github.com/OpenCyphal/CETL/blob/4c754a5e58dcb88e06aef89ddb08c01876e07a18/include/cetl/unbounded_variant.hpp#L1109

https://github.com/OpenCyphal/CETL/blob/4c754a5e58dcb88e06aef89ddb08c01876e07a18/include/cetl/unbounded_variant.hpp#L1130

https://github.com/OpenCyphal/CETL/blob/4c754a5e58dcb88e06aef89ddb08c01876e07a18/include/cetl/unbounded_variant.hpp#L1156

https://github.com/OpenCyphal/CETL/blob/4c754a5e58dcb88e06aef89ddb08c01876e07a18/include/cetl/unbounded_variant.hpp#L1180

So that one does not have to touch the ad-hoc `in_place_type`

serges147:
> So that one does not have to touch the ad-hoc `in_place_type`

Good idea, I will do this.

pavel-kirienko (approved):

thirtytwobits (approved):

PR #127, opened 2024-06-05
Added `type_id()` & `type_size()` methods to the `cetl::unbounded_variant`. #verification #docs #sonar

serges147:


include/cetl/unbounded_variant.hpp
```diff
@@ -556,27 +557,55 @@ struct base_access : base_storage<Pmr, Footprint, Alignment>
...
+        return value_const_converter_(base::get_raw_storage(), {}).second;
+    }
+
+    /// \brief Returns the size of the stored value in bytes.
+    ///        Zero if storage is empty.
+    ///
+    CETL_NODISCARD std::size_t type_size() const noexcept
+    {
+        return has_value() ? base::get_value_size() : 0UL;
+    }
```
serges147, line 607: Both these methods are very useful at least at libcyphal unit tests. But both of them could be useful also at production code - to have a peek what value (or size) is currently stored at ub_var.

serges147: Also, note that there is no size penalty for these two - size is already these, and querying of type id is done by reusing existing const convertor.

include/cetl/unbounded_variant.hpp
```diff
@@ -588,7 +617,7 @@ struct base_access : base_storage<Pmr, Footprint, Alignment>
         }
         CETL_DEBUG_ASSERT(nullptr != value_const_converter_, "Non-empty storage is expected to have value converter.");
 
-        return value_converter_(base::get_raw_storage(), type_id_value<ValueType>);
+        return value_mut_converter_(base::get_raw_storage(), cetl::type_id_value<ValueType>);
```
serges147, line 620: Renamed to emphasise that there are two distinct converts: mutable and constant.

cetlvast/suites/unittest/test_unbounded_variant.cpp
```diff
@@ -19,6 +19,44 @@
 
 // NOLINTBEGIN(*-use-after-move)
 
+namespace cetl
```
pavel-kirienko, line 22: Maybe these should be moved into a header to allow reuse in other tests, shall the need arise?

include/cetl/rtti.hpp
```diff
@@ -81,6 +82,15 @@ constexpr type_id type_id_type_value() noexcept
...
 
+/// The type ID value specializations for `void` type.
+///
+/// In use for the `type_id_invalid` constant, which f.e. is in use at
+/// `unbounded_variant::type_id()` method to indicate that the variant is valueless.
+///
+template <>
+constexpr type_id type_id_value<void>{};
+constexpr type_id type_id_invalid{type_id_value<void>};
+
```
pavel-kirienko, line 93: This is not good because it breaks the orthogonality of the types. All type-ID values should be valid; if you need to represent the absence of a type-ID, use `optional`. Seeing as we can't eat our own cookies, we can define a special case for `unbounded_variant::type_id` such that it returns the type-ID of `void` if it's empty.

Essentially I'm saying that we should keep your current design but remove this special `type_id_invalid` and confine the poor typing to one method because there's no other way.

include/cetl/rtti.hpp
```diff
@@ -35,6 +35,7 @@ using type_id = std::array<std::uint8_t, type_id_size>;
 /// The bytes of the UUID are given as a list of template parameters; there shall be at most 16 of them;
 /// if any are missing, they are assumed to be 0.
 /// For conversion to \ref type_id use \ref cetl::type_id_type_value.
+/// Please don't use empty or all zeros bytes, as it will be treated as the invalid type ID (see \ref type_id_invalid).
```
pavel-kirienko, line 38:
```suggestion
/// Please don't use empty or all zeros bytes.
```

It's a good advise regardless of the issue below.

pavel-kirienko (approved):
After the special case is removed it should be good to go

cetlvast/suites/unittest/test_unbounded_variant.cpp
serges147: ok, I'll create helper header.

include/cetl/rtti.hpp
pavel-kirienko: image

cetlvast/suites/compile/test_unbounded_variant_footprint_get_const.cpp
```diff
@@ -8,19 +8,10 @@
...
 
-namespace cetl
-{
-template <>
-constexpr type_id type_id_value<uint8_t>{};
-
-template <>
-constexpr type_id type_id_value<uint16_t>{};
-
-}  // namespace cetl
```
serges147, line 22: Note that in examples I deliberately don't use this new `cetlvast/helpers_rtti.hpp` and I left such definitions as is. Rationale: so that examples will be more self-sufficient and explicitly show how to declare a type id.

pavel-kirienko (approved):

pavel-kirienko (approved):

include/cetl/unbounded_variant.hpp
```diff
@@ -556,27 +557,55 @@ struct base_access : base_storage<Pmr, Footprint, Alignment>
     template <typename Tp, std::enable_if_t<is_rtti_convertible<Tp>, int> = 0>
     void make_converters() noexcept
     {
-        value_const_converter_ = [](const void* const storage, const type_id& id) {
-            const auto ptr = static_cast<const Tp*>(storage);
-            return ptr->_cast_(id);
+        value_const_converter_ = [](const void* const storage, const cetl::type_id& dst_type_id) {
```
thirtytwobits, line 560: These lambdas may fail on systems with no heap and a deficient small-value-optimization in std::function. Please use function pointers to avoid this.

serges147: I don't think so, they are just pure (implicitly generated) static functions, they don't capture anything (neither any of outside vars nor `this` - so no allocation. sizeof of them is 1 (should kinda be zero but as you know even empty struct sizeof is 1). And the fact that I can just assign them to a simple pointer to function (see `value_const_converter_` declaration below) proves it - as soon as add anything to the capture list such assignments won't work anymore.

Here is sizeof of this lambda:
```
        using Xxx = decltype([](const void* const storage, const cetl::type_id& dst_type_id) {
            CETL_DEBUG_ASSERT(nullptr != storage, "");
            const auto        ptr     = static_cast<const Tp*>(storage);
            const void* const dst_ptr = ptr->_cast_(dst_type_id);
            return std::make_pair(dst_ptr, cetl::type_id_value<Tp>);
        });
        static_assert(sizeof(Xxx) == 1);
```

serges147: In addition, although I could technically make explicit template static member functions for convertors, I'm reluctant todo so b/c:
1. There will be 4 of them (see two mutually exclusive versions of `make_converters`).
2. There all will be out of `make_converters` context
3. `make_converters` is the only supposed entity to reference them; and of course nobody is allowed to call them directly (only indirectly via `value_[mut|const]_convertor_` function pointers). So, in some sense I believe am doing proper isolation here without introducing any indeterministic behavior.

serges147: Furthermore, I scanned AUTOSAR14 spec, and found no lambda rule against my implementation. The only remotely applicable rule is the following:
>Rule A5-1-6 (advisory, implementation, automated)
Return type of a non-void return type lambda expression should be
explicitly specified.

I can fix it.

serges147: done - see [commit](https://github.com/OpenCyphal/CETL/pull/127/commits/e0adc23126e5c014cdd295c5295d3bf5887fa4ce)

pavel-kirienko: To back up what Sergei said:

image

pavel-kirienko (dismissed):

include/cetl/rtti.hpp
```diff
@@ -75,11 +76,28 @@ constexpr type_id type_id_type_value() noexcept
     return detail::type_id_type_value_impl(TypeIDType{});
 }
 
+/// The type ID getter for the given type.
+/// This helper is provided for regularity; it returns the same value as \c T::_get_type_id_().
+/// The type shall satisfy \ref cetl::has_type_id.
```
pavel-kirienko, line 81:
```suggestion
/// The type shall satisfy \ref cetl::has_type_id.
/// Specialize this getter to add RTTI support to types where it is not possible to define a static method 
/// (e.g., builtins, pointers, third-party classes, etc).
```

pavel-kirienko (approved):

pavel-kirienko (approved):

thirtytwobits (approved):
Looks good. Thanks.

PR #142, opened 2024-10-17
First draft of `cetl::pf17::string_view`

serges147:
- First pr for issue #2 . Not all c++17 overloads are implemented but most of basic ones are there.

### DIFF
--- a/cetlvast/suites/unittest/test_pf17_string_view.cpp
+++ b/cetlvast/suites/unittest/test_pf17_string_view.cpp
@@ -15,6 +15,8 @@
 #include <array>
 #include <cassert>
 #include <cstdlib>
+#include <iomanip>
+#include <sstream>
 #include <string>
 #include <cetl/cetl.hpp>
 
@@ -54,6 +56,7 @@ public:
         std::array<wchar_t, 256> wstr{};
         const auto result = std::mbstowcs(wstr.data(), cstr, wstr.size());
         assert(result == std::strlen(cstr));
+        (void) result;
         return {wstr.data()};
     }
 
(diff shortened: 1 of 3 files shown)


include/cetl/pf17/string_view.hpp
```diff
@@ -0,0 +1,295 @@
...
+    using const_pointer   = const char*;
+    using reference       = const char&;
+    using const_reference = const char&;
+    using const_iterator  = const char*;
+    using iterator        = const_iterator;
+    using size_type       = std::size_t;
+
+    enum : size_type
+    {
+        npos = static_cast<size_type>(-1)
```
pavel-kirienko, line 38: Why not static constexpr?

pavel-kirienko (approved):
Are the docs omitted on purpose?

cetlvast/suites/unittest/test_pf17_string_view.cpp
```diff
@@ -0,0 +1,323 @@
+/// @file
```
thirtytwobits, line 1: It's critical that we start with type-parameterized tests that compare the CETL implementation to the ones in our representative standard libraries like https://github.com/OpenCyphal/CETL/blob/main/cetlvast/suites/unittest/test_pf20_span.cpp does for span.

include/cetl/pf17/string_view.hpp
```diff
@@ -0,0 +1,295 @@
...
+#include <exception>  // We need this even if exceptions are disabled for std::terminate.
+#include <stdexcept>
+#include <string>
+
+namespace cetl
+{
+namespace pf17
+{
+
+class string_view
```
thirtytwobits, line 23: In the standard, string_view is just an alias for basic_string_view<char> so we need to start with this template.

thirtytwobits (changes requested):

include/cetl/pf17/string_view.hpp
serges147: c++14

serges147:
> Are the docs omitted on purpose?

You right, I for some reason thought that we don't duplicate docs (cppreference) for our polyfills, but now I see that span, variant and etc do contain certain level of documentation - I'll fix it.

include/cetl/pf17/string_view.hpp
serges147: ok, I can do it. My thinking was that it will add more work while we most probably will use `char` specialization only.

cetlvast/suites/unittest/test_pf17_string_view.cpp
serges147: 1. See below my response, namely:
>My thinking was that it will add more work while we most probably will use char specialization only.

Could you please elaborate why it is critical to invest into what we most probably (plz correct me if I'm wrong)  will no use? Are we making CETL universal for the whole world? Honestly, it's quite exhausting to "fight" against c++14 - I believe libcyphal were long time done already if it was not support of 14 standard. And Nunavut's c++ code generation goals will be much simpler and more clean. I do understand that anyway there will be always a "race" against the latest standard and what we "have to support", so polyfills will probably still live in some form... it just concerns me (a lot!) that a lot of simple things become not so simple, and hardly contribute to the final product we deliver to customers. Sorry, for my frustration and being emotional ;)

2. BTW, you said 
>... compare the CETL implementation to the ones in our representative standard libraries ..."

I didn't quite catch what you meant by "compare"... I don't see there any comparison of what `cetl::pf20::span` and `std::span` does... if this is what you meant.

thirtytwobits: The simple answer is that CETL's polyfill types are re-implementations from the standard and the standard defines basic_stringview then provides aliases. We can tolerate sparse implementations but must minimize deviant implementations.

As for the reasoning for C++14? It just goes to the state of the world and compatibility. The need for C++14 polyfill types will naturally diminish over time just as all software gets outdated and deprecated as the years roll by.

thirtytwobits (changes requested):
This is fine but please add the A/B comparison test between STL and CETL. See test_variable_length_array_compat.cpp:159 "Any Type" for how to swap between the two in parameterized tests.

thirtytwobits:
Well, that didn't go as I had planned. I pushed an example of how to write these A/B tests but it ended up "merging" your PR into this repo's new string_view_2 branch. Can you take a look and finish your work up on that branch or pull it over and reopen this one? Sorry.

PR #146, opened 2024-11-08
Use latest gtest v1.15.2 and O1Heap

serges147:


serges147:
Can't b/c of getting errors:
```
error: use of the 'maybe_unused' attribute is a C++17 extension [-Werror,-Wc++17-attribute-extensions]
```

Issue #164, opened 2025-02-27
cetlvast: stringop-overflow warning, gcc, c++20, libstdc++


thirtytwobits:
Building test_variable_length_array_compat.cpp using gcc --std=c++20 yields the below failure. When using clang --std=c++20 -stdlib=libc++ this warning does not occur.
```
[build] FAILED: suites/unittest/CMakeFiles/test_variable_length_array_compat__googletest_objlib.dir/Release/test_variable_length_array_compat.cpp.o 
[build] /usr/bin/g++ -DCETL_ENABLE_DEBUG_ASSERT=0 -DCETL_VERSION=\"0.0.0\" -DLIBCXX_ENABLE_ASSERTIONS=0 -DCMAKE_INTDIR=\"Release\" -I/repo/cetlvast/include -I/repo/include -I/repo/cetlvast/build_external/o1heap/o1heap -isystem /repo/cetlvast/build_external/googletest/googletest/include -isystem /repo/cetlvast/build_external/googletest/googletest -isystem /repo/cetlvast/build_external/googletest/googlemock/include -isystem /repo/cetlvast/build_external/googletest/googlemock -O3 -DNDEBUG -std=c++20 -pedantic -Wall -Wextra -Werror -Wfloat-equal -Wconversion -Wunused-parameter -Wunused-variable -Wunused-value -Wcast-align -Wmissing-declarations -Wmissing-field-initializers -Wdouble-promotion -Wswitch-enum -Wtype-limits -Wno-error=array-bounds -O3 -fno-delete-null-pointer-checks -Wsign-conversion -Wsign-promo -Wold-style-cast -Wzero-as-null-pointer-constant -Wnon-virtual-dtor -Woverloaded-virtual -MD -MT suites/unittest/CMakeFiles/test_variable_length_array_compat__googletest_objlib.dir/Release/test_variable_length_array_compat.cpp.o -MF suites/unittest/CMakeFiles/test_variable_length_array_compat__googletest_objlib.dir/Release/test_variable_length_array_compat.cpp.o.d -o suites/unittest/CMakeFiles/test_variable_length_array_compat__googletest_objlib.dir/Release/test_variable_length_array_compat.cpp.o -c /repo/cetlvast/suites/unittest/test_variable_length_array_compat.cpp
[build] In file included from /usr/include/c++/13/bits/stl_iterator.h:85,
[build]                  from /usr/include/c++/13/bits/stl_algobase.h:67,
[...]

Issue #160, opened 2025-02-10
use cmake presets


thirtytwobits:
verify.py is unnecessary now that cmake has presets. Use this instead and get rid of verify.py

PR #167, opened 2025-03-04
Adding cmake presets, switching to ninja-multi-config

thirtytwobits:
Features:

- Switch to cmake presets (#160) and removal of custom python execution wrapper
- Properly uses `libc++` on clang and `libstdc++` on gcc (fixes for libc++ changes included)
- Formatter updates
- GCC7 ready (still need to fix the bugs but we can build for it now)
- builds on arm without timing out.
- fix for #164


cetlvast/suites/unittest/CMakeLists.txt
```diff
@@ -86,19 +190,15 @@ set_directory_properties(PROPERTIES
...
-if (CMAKE_BUILD_TYPE STREQUAL "Coverage")
-    enable_coverage_report(COVERAGE_REPORT_FORMATS html sonarqube
-                           ROOT_DIRECTORY ${CETL_ROOT}
-                           OUT_REPORT_INDICES LOCAL_COVERAGE_REPORT_INDICIES
-    )
-endif()
+enable_coverage_report(COVERAGE_REPORT_FORMATS html sonarqube
+                        ROOT_DIRECTORY ${CETL_ROOT}
+                        OUT_REPORT_INDICES LOCAL_COVERAGE_REPORT_INDICIES
+)
```
serges147, line 196: I ran "configure-gcc-native-cpp-14-offline - build-Coverage-gcc-native-cpp-14-offline" and opened `build/suites/unittest/gcovr_html/coverage.html`, and noticed quite bad % of Branches coverage.
image
I believe this is heavily affected by our `CETL_DEBUG_ASSERT`, like here:
image

I propose we suppress CETL_DEBUG_ASSERT for Coverage builds.

serges147 (approved):
Please see my comments. Thank You!

cetlvast/CMakePresetsVendorTemplate.json
```diff
@@ -0,0 +1,159 @@
...
+        "minor": 30,
+        "patch": 0
+    },
+    "configurePresets": [],
+    "buildPresets": [],
+    "workflowPresets": [],
+    "packagePresets": [],
+    "testPresets": [],
+    "vendor": {
+        "tcpm": {
```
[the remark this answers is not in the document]
thirtytwobits: Sorry, yeah, this is an open-source utility I published a couple weeks ago:

https://pypi.org/project/tcpm/

It's part of a conversation we're having in the cmake community:

https://gitlab.kitware.com/cmake/cmake/-/issues/22538

cetlvast/CMakePresets.json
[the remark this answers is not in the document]
thirtytwobits: Vscode is actually really good about only showing valid options. It's too bad CLion doesn't do this as well. As mentioned below, [this is an ongoing area of discussion in the cmake community](https://gitlab.kitware.com/cmake/cmake/-/issues/22538)

cetlvast/suites/unittest/CMakeLists.txt
thirtytwobits: Should we not just try to get the coverage numbers up over time?

serges147: Numbers are already good (see **Lines** and even **Functions** coverage types). Here I'm specifically talking about **Branches** coverage type numbers - these should be also not  bad, but with non-empty `CETL_DEBUG_ASSERT` it won't be possible improve them. It's very rarely you could artificially craft environment so, that assert will be failing (and caught by "expected to crash" test). IMHO targetting asserts coverage is too much. But leaving them as is means that we won't detect real lacks of branch coverage in some production business logic. In other words - it's noise for me.

thirtytwobits: ok

serges147 (dismissed):

PR #183, opened 2025-03-25
Working around GCC7 bug handling temporary template parameters

thirtytwobits:
- fixes issue/151
- a couple more instances for issues/120
- removing some invalid C++23 cmake presets
- removing gcc7 ifdef around variant constexpr

This last one is a mystery. It appears the GCC7 constexpr bug was being caused by something else that has since been fixed as we've worked around better understood GCC7 bugs.

### DIFF
--- a/cetlvast/CMakePresets.json
+++ b/cetlvast/CMakePresets.json
@@ -295,24 +295,6 @@
             ]
         },
         {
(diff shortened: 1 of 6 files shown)


pavel-kirienko (approved):
not sure if removing the `Alt` alias helped with readability though

Issue #149, opened 2025-01-18
Add builds for Arduino (atmelsam)


thirtytwobits:
CETL has some value for allowing more modern C++ on old but popular Arduino toolchains like GCC 7.2.1. We should support it to increase adoption and contributions.

pavel-kirienko:
the amount of compiler bugs requiring workarounds is mildly disturbing

PR #184, opened 2025-03-25
Progress on #149 Arduino/Embedded

thirtytwobits:
A small step towards testing CETL on MCU hardware.

### DIFF
--- a/.devcontainer/cmake/toolchains/arm-none-eabi-gcc.cmake
+++ b/.devcontainer/cmake/toolchains/arm-none-eabi-gcc.cmake
@@ -1,0 +1,47 @@
+#
+# Copyright (C) OpenCyphal Development Team  <opencyphal.org>
+# Copyright Amazon.com Inc. or its affiliates.
+# SPDX-License-Identifier: MIT
+#
+
+# armv7m cross-compilation
+#
+set(CMAKE_C_COMPILER arm-none-eabi-gcc CACHE FILEPATH "C compiler")
+set(CMAKE_CXX_COMPILER  arm-none-eabi-g++ CACHE FILEPATH "C++ compiler")
+set(CMAKE_ASM_COMPILER  arm-none-eabi-gcc CACHE FILEPATH "assembler")
+
+set(CMAKE_SYSTEM_NAME Generic)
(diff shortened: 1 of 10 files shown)


.devcontainer/cmake/toolchains/arm-none-eabi-gcc.cmake
```diff
@@ -0,0 +1,47 @@
...
+set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+
+
+# This are compile options for all modules we compile. Architecture-specific stuff should be inherited from the
+# board library's compile options since it's that module that defines what CPU we're running on.
+add_compile_options(
+    -fdata-sections
+    -ffunction-sections
+    -fverbose-asm
+    -mcpu=cortex-m7
```
pavel-kirienko, line 33: Isn't this supposed to be set from `CMAKE_SYSTEM_PROCESSOR` by CMake?

pavel-kirienko (approved):

.devcontainer/cmake/toolchains/arm-none-eabi-gcc.cmake
thirtytwobits: This doesn't seem to work based on my experimentation. The documentation for `CMAKE_SYSTEM_PROCESSOR` reads:

> When cross-compiling, a [CMAKE_TOOLCHAIN_FILE](https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html#variable:CMAKE_TOOLCHAIN_FILE) should set the CMAKE_SYSTEM_PROCESSOR variable to match target architecture that it specifies (via [CMAKE_\<LANG\>_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) and perhaps [CMAKE_\<LANG\>_COMPILER_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_TARGET.html#variable:CMAKE_%3CLANG%3E_COMPILER_TARGET)).

which suggests one still has to set the compiler flag manually.

PR #185, opened 2025-04-10
Fix for issue/141: `cetl::get_if` should should match only on a `unbounded_variant`

serges147:


include/cetl/cetl.hpp
```diff
@@ -128,12 +128,13 @@
...
+// Intentional violation of Sonar: the assertions check macro cannot be replaced with a function definition.
+#if defined CETL_ENABLE_DEBUG_ASSERT && CETL_ENABLE_DEBUG_ASSERT
 #    include <cassert>
-#    define CETL_DEBUG_ASSERT(c, m) assert(((void) (m), (c)))
+#    define CETL_DEBUG_ASSERT(c, m) assert(((void) (m), (c)))  // NOSONAR cpp:S960
 #else
-#    define CETL_DEBUG_ASSERT(c, m) ((void) (m))
-#endif  // CETL_ENABLE_DEBUG_ASSERT
+#    define CETL_DEBUG_ASSERT(c, m) ((void) (m))  // NOSONAR cpp:S960
+#endif
```
serges147, line 137: `CETL_ENABLE_DEBUG_ASSERT` now always defined, either to `0` or `1`.

include/cetl/unbounded_variant.hpp
```diff
@@ -1615,122 +1627,160 @@ CETL_NODISCARD UnboundedVariant make_unbounded_variant(typename UnboundedVariant
     return UnboundedVariant(mem_res, in_place_type_t{}, list, std::forward<Args>(args)...);
 }
 
-/// \brief Performs type-safe access to the contained object.
```
serges147, line 1618: It's hard to compare b/c I have to swap order of `get`-s and `get_if`-s (otherwise clang fails (gcc was okey)).
In reality there little changes. The most important are:
- `get`-s (const and non-const) now return `ValueType` reference - similar to `variant::get`-s. Previously it was a value copy, which was not only non-consistent with `variant` but also too restrictive for "copyable" types only. Now such limitation is lifted.
- usage of `std::enable_if_t` - actual fix for the issue.

pavel-kirienko (dismissed):

cetlvast/suites/unittest/test_unbounded_variant.cpp
```diff
@@ -458,10 +458,10 @@ TEST_F(TestPmrUnboundedVariant, ctor_2_copy)
         const ub_var src{test{}};
         ub_var       dst{src};
 
-        EXPECT_THAT(get<test>(src).value_, 1 + 10);
+        EXPECT_THAT(get<test>(src).value_, 1);
         EXPECT_THAT(get<const test&>(src).value_, 1);
```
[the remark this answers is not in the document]
serges147: Previously `get` function always returned copy of a value (hence extra `+ 10`). Now method returns value reference (const or non-const depending on constness of the input unbounded variant).

Issue #187, opened 2025-07-30
Memory Leak in VariableLengthArrayBase Move Constructor with Allocator When `is_pocma_or_is_always_equal<Alloc>` is `false_type`


JLReitz:
See relevant test case with failing results below:
```c++
using VLAWithNoPropagateOnMoveAllocator =
    cetl::VariableLengthArray<
            InstrumentedType,
            cetlvast::InstrumentedNewDeleteAllocator<
                InstrumentedType,
                std::false_type,  // Is always equal
                std::false_type,  // Is equal
                std::false_type,  // Propagate on container move assignment
                std::true_type    // Propagate on container copy assignment
            >
        >;
using VLAWithNoPropagateOnMoveAllocatorTests = VLADetailedAllocationTests<VLAWithNoPropagateOnMoveAllocator>;
TEST_F(VLAWithNoPropagateOnMoveAllocatorTests, MoveConstructWithNewAllocator)
{
    using TestFixture = VLAWithNoPropagateOnMoveAllocatorTests;
    using VlaType = VLAWithNoPropagateOnMoveAllocator;
    using AllocatorType = VlaType::allocator_type;

    VlaType test_source{{1, 2, 3, 4}, AllocatorType{}};
    VlaType test_subject{std::move(test_source), AllocatorType{}};

    EXPECT_EQ(12, TestFixture::ItemT::total_instances_constructed);
    EXPECT_EQ(4, TestFixture::ItemT::total_instances_implicit_int_constructed); // Test source init list instances
    EXPECT_EQ(4, TestFixture::ItemT::total_instances_copy_constructed); // Test source instances
    EXPECT_EQ(4, TestFixture::ItemT::total_instances_move_constructed); // Test subject instances
    EXPECT_EQ(2, cetlvast::InstrumentedAllocatorStatistics::get().allocations);
    EXPECT_EQ(2, cetlvast::InstrumentedAllocatorStatistics::get().deallocations);
}
```
```
[----------] 1 test from VLAWithNonMovePropagationAllocatorTests
[ RUN      ] VLAWithNonMovePropagationAllocatorTests.MoveConstructWithNewAllocator
/repo/cetlvast/suites/unittest/test_variable_length_array_detailed_allocation.cpp:754: Failure
Expected equality of these values:
  2
  cetlvast::InstrumentedAllocatorStatistics::get().deallocations
    Which is: 1

[...]

PR #188, opened 2025-07-30
Fix for #187

JLReitz:
`VariableLengthArrayBase` move construction with allocator, when `is_pocma_or_is_always_equal<Alloc>` is `false_type` and `this->alloc_ != rhs.alloc_`, will deallocate `rhs.data_` prior to setting `rhs.data_ = nullptr`.


thirtytwobits:
👀

JLReitz:
Prior to this change deallocation would not happen in the moved-from object's destructor because it's `data_` member is set to `nullptr` in the moved-to object's move constructor with allocator. See #187.

To verify you may comment out line 848 of include/cetl/variable_length_array.hpp and run the CETLVast suite, expect the new unit test I've added to fail at the deallocation count expectation.

thirtytwobits (approved):

